PR #11, opened 2024-05-21
Feature/signer

AusIV:


natefikru:
curious as to why some of my unit test changes are showing up as changes in this PR. hmm

auth/interfaces.go
```diff
@@ -0,0 +1,77 @@
...
+	if at.Uses != nil && atomic.LoadInt64(at.Uses) <= 0 {
+		return ErrRequestLimit
+	}
+	if at.Expiration != nil && time.Since(time.Time(*at.Expiration)) > 0 {
+		return ErrSessionExpired
+	}
+	return nil
+}
+
+type AuthClient interface {
```
natefikru, line 65: a nice to have would be prefixing this interface with "I" as in "IAuthClient" so it can be easily determined that this is an interface by reading the struct type.

auth/interfaces.go
```diff
@@ -0,0 +1,77 @@
...
+type UnixTime time.Time
+
+var (
+	ErrRequestLimit = errors.New("request limit exceeded")
+	ErrSessionExpired = errors.New("session expired")
+	ErrSessionClosed = errors.New("session closed")
+	ErrNoTokensAvailable = errors.New("no tokens available")
+)
+
+func (t UnixTime) MarshalJSON() ([]byte, error) {
```
natefikru, line 20: this file could possibly be split up into 2, one file for the interface definitions, the other labeled helpers.go or utils.go or something of the like, since most of the code blocks in this file doesn't have anything to do with interfaces.

natefikru: AuthToken could also have its own file.

auth/eip4361/client.go
```diff
@@ -0,0 +1,269 @@
+package eip4361
```
natefikru, line 1: would it make sense to put the /auth folder within the /lib directory? up to you.

auth/eip4361/client.go
```diff
@@ -0,0 +1,269 @@
...
+	"time"
+	"github.com/spruceid/siwe-go"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/crypto"
+	"hash/fnv"
+	"go.uber.org/zap"
+)
+
+type SigningConfig struct {
```
natefikru, line 23: SigningConfig can also probably have its own file within the eip4361 directory.

auth/eip4361/client.go
```diff
@@ -0,0 +1,269 @@
...
+	return auth.ErrNoTokensAvailable
+}
+
+type signedMessage struct {
+	Message string `json:"msg"`
+	Signature hexutil.Bytes `json:"sig"`
+}
+
+// GetToken should take a map of parameters for a token, and return a map of Header -> Value for a session
+func (c *EIP4361ClientAuth) GetToken(map[string]interface{}) (auth.AuthToken, error) {
```
natefikru, line 194: would it make sense to return back the pointer for auth.AuthToken here?

natefikru:
I appreciate all of the comments you put in explaining the functions 🙏

module.go
```diff
@@ -5,15 +5,19 @@ import (
...
 	prom "github.com/openrelayxyz/din-caddy-plugins/lib/prometheus"
 	mod "github.com/openrelayxyz/din-caddy-plugins/modules"
+	"github.com/openrelayxyz/din-caddy-plugins/auth/eip4361"
 )
 
 func init() {
 	caddy.RegisterModule(mod.DinUpstreams{})
 	caddy.RegisterModule(mod.DinSelect{})
 	caddy.RegisterModule(mod.DinMiddleware{})
+	caddy.RegisterModule(eip4361.EIP4361AuthMiddleware{})
```
natefikru, line 15: I'm assuming since this is still a draft, the middleware module for auth hasn't been created yet.

PR #14, opened 2024-05-29
Bugfix/healthchecks

AusIV:


modules/din_upstreams.go
```diff
@@ -20,11 +24,92 @@ var (
...
+	successes int
+	healthy   bool
+	quit      chan struct{}
+}
+
+func (u *upstreamWrapper) Available() bool {
+	return u.upstream.Available() && u.Healthy()
+}
+
+func (u *upstreamWrapper) StartHealthchecks() {
```
natefikru, line 43: would be nice to have a couple comments going through the health check process.

Caddyfile
```diff
@@ -120,13 +120,17 @@
 				starknet-sepolia {
 					methods web3_sha3 web3_clientVersion net_listening net_peerCount net_version eth_call eth_getBalance eth_estimateGas eth_createAccessList eth_getStorageAt eth_getCode eth_blockNumber eth_protocolVersion eth_syncing eth_sendRawTransaction eth_chainId eth_getLogs eth_getTransactionByHash eth_getTransactionReceipt eth_getTransactionCount eth_feeHistory eth_getBlockByNumber eth_getBlockByHash eth_gasPrice eth_getTransactionByBlockHashAndIndex eth_getTransactionByBlockNumberAndIndex eth_getBlockTransactionCountByNumber eth_getBlockTransactionCountByHash eth_getUncleCountByBlockNumber eth_getUncleCountByBlockHash eth_subscribe eth_unsubscribe eth_getUncleByBlockHashAndIndex eth_maxPriorityFeePerGas eth_getProof
 					providers {
-						https://starknet-sepolia.blastapi.io:443/key
+						https://starknet-sepolia.blastapi.io:443/key {
+							healthcheck_method starknet_blockNumber
```
natefikru, line 124: is the request actually `starknet_blockNumber` and not `eth_blockNumber`

modules/din_upstreams.go
```diff
@@ -20,11 +24,92 @@ var (
...
+		for i :=0 ; ; i++ {
+			select {
+			case <-u.quit:
+				ticker.Stop()
+				return
+			case <-ticker.C:
+				rpcCallString := fmt.Sprintf(`{"jsonrpc":"2.0","id":%v,"method":"%v","params":[]}`, i, u.HCRPCMethod)
+				req, err := http.NewRequest("POST", u.HttpUrl, strings.NewReader(rpcCallString))
+				if err != nil {
+					u.markFailure(err.Error())
```
natefikru, line 69: i feel like this wouldn't be a criteria to mark the proxy for error but instead to send an error via a channel or something. I'm not sure. It probably isn't that big of a deal though.

modules/din_upstreams.go
```diff
@@ -20,11 +24,92 @@ var (
...
+				u.markSuccess()
+			}
+		}
+	}()
+}
+
+func (u *upstreamWrapper) markFailure(reason string) {
+	u.failures++
+	u.successes = 0
+	if u.failures > 2 && u.healthy {
```
natefikru, line 97: should this 2 value actually be a const?

modules/din_upstreams.go
natefikru, line 43: we probably should have a unit test here but i don't want to delay deployment for this.

natefikru:
It looks like there is a unit testing failing, however, overall there are no major issues that I've seen as long as you've tested and made sure that this works proper, so I will approve. I just have some minor comments here and there.

natefikru (approved):

modules/din_upstreams.go
```diff
@@ -20,11 +24,92 @@ var (
...
+		u.healthy = false
+		// fmt.Printf("Marking %v as unhealthy: %v\n", u.HttpUrl, reason)
+	}
+}
+func (u *upstreamWrapper) markSuccess() {
+	u.successes++
+	if u.successes > 2 && !u.healthy {
```
egalano, line 104: Same comment as nate's I think it would be good to make the healthy and unhealthy threshold a constant we can quickly find and modify.

egalano (approved):
LGTM pending

Caddyfile
AusIV: Yes. Starknet endpoints will return an error status on eth_blockNumber, but return successfully with starknet_blockNumber

modules/din_upstreams.go
AusIV: Looking at https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/net/http/request.go;l=855

The only way NewRequest returns an error is if: 

* The method is invalid (but "POST" is hard coded and always valid)
* The URL is invalid (but u.HttpUrl was validated in the middleware's unmarshaller)

Since neither of these conditions can be met given the current codebase, I could make a case for ignoring the error entirely. My concern is that as we start to integrate with the smart contract we may not do as much validation of the URL, and at the very least having the upstream get marked as unhealthy will likely save us some grief.

Once we integrate with the authentication stuff (which passes logging around) we'll have logs of the errors, so if something is unhealthy because of an invalid URL, that should become reasonably obvious from the logs that markFailure will emit.

AusIV:
I went ahead and made the constant values configurable. That had been my intent from the outset, but forgot to loop back around to it after I got things working.

modules/din_upstreams.go
natefikru: good point. i think in this case you're right that marking it as unhealthy is probably better than just throwing an error.

Caddyfile
natefikru: wow what a peculiar network.

PR #19, opened 2024-06-07
Feature/add block health consistency check

natefikru:
built on top of the [ping health check bugfix PR](https://github.com/openrelayxyz/din-caddy-plugins/pull/14) written by @AusIV 

**Changes include**

**Block health consistency health check** 
- This update checks all provider's associated with a service every "x" amount of seconds and checks to see if the provider is serving the latest blockNumber in contrast with the other supported providers. If a `new` latest blockNumber is found, mark that provider as healthy and all of the `old` blockNumber providers as unhealthy. This will ensure that only the providers with the latest blockNumber values are marked as healthy.

**Refactor/Variable changes**
- renamed upstreamWrapper struct to provider, This is to more logically match the naming scheme of the CaddyFile and reduce naming confusion.
- abstracted service and provider structs into their own files in the modules package
- changed the middleware.Services data structure. This is also done to better match the CaddyFile definition setup. Now instead of a list of services with a list of providers(upstreamWrappers), we now have a map of services that houses a map of providers. 
- The a service within the middleware.Services map also now houses the data structure for the list of methods defined in the Caddyfile.

**Tests**
- Updated existing tests to support updated middleware data structure 
- Added tests for health check function

**Update (June 25th)**
- Added default blockLag buffer value that gives a cushion for block number to be behind. Can also be set statically in the Caddyfile/service config.
- when a provider's blocknumber is checked, the latest blocknumber is stored at index[0] of the provider's healthCheck list, and only 10 of the latest healthChecks are stored along with the timestamp.
- Added Healthy, Warning and Unhealthy Enum/iota values for health status'
- Added filtering of upstreams based on warning status. if no healthy/warning upstreams are returned, then no upstreams are returned for the request and the request fails.


modules/service.go
```diff
@@ -0,0 +1,110 @@
...
+		// get the latest block number from the current provider
+		providerBlockNumber, statusCode, err := s.runtimeClient.GetLatestBlockNumber(provider.HttpUrl, provider.Headers)
+		if err != nil {
+			fmt.Println(err, "Error getting latest block number for provider", provider.host, "on service", s.Name)
+			provider.markPingFailure(s.HCThreshold)
+			continue
+		}
+
+		// Ping Health Check
+		if statusCode > 399 {
```
egalano, line 60: Based on past experience, it would be good to associate the latest block number with a timestamp. Because if the blocknumber does not change within a certain amount of time, the backend is considered unhealthy/stalled. This time should be configurable per network/service.

modules/din_upstreams_test.go
```diff
@@ -49,72 +49,74 @@ func TestGetDinUpstreams(t *testing.T) {
...
-				{
+			replacerProviders: map[string]*provider{
+				upstream1.Dial: {
 					upstream: upstream1,
 					Priority: 0,
 				},
-				{
+				upstream2.Dial: {
 					upstream: upstream2,
 					Priority: 0,
```
egalano, line 67: Do we have a test to verify that priority is being adhered to now? We ran into a bug with this the other week

egalano (approved):
Generally LGTM Nate. I was mainly looking at the logic of the health checks since I cant scrutinize the go style too closely. Only question I have on that is if there is an alternative way to structure the health checks so its easier to extend or augment them going forward to make them more robust. For instance tracking latest 10 block numbers per provider rather than just the last one, and making that code reusable externally when we separate this functionality from the router itself.

modules/service.go
AusIV: This line is looking at HTTP response codes, not block numbers. In general I'd expect healthy response codes to be 200, but if it's in the 4xx-5xx range it's definitely got a problem.

modules/service.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+		// add the current provider to the checked providers map
+		checkedProviders[provider.upstream.Dial] = providerBlockNumber
+	}
+}
+
+func (s *service) evaluateCheckedProviders(checkedProviders map[string]int64) {
+	for providerName, blockNumber := range checkedProviders {
+		if blockNumber < s.LatestBlockNumber {
+			s.Providers[providerName].healthy = false
```
AusIV, line 92: I'm not sure I like using `healthy` for both totally unresponsive servers and block numbers running behind.

If everything is unhealthy, we'll attempt to load balance the request across all of the servers. Servers that are unhealthy because they haven't updated their block numbers in a while are probably better candidates for routing than servers that are unhealthy because they're unroutable or serving 5xx errors.

A Rivet we use Consul as a service mesh, and Consul has three statuses - Healthy, Unhealthy, and Warning. Healthy / Unhealthy do what you expect, and Warning status is typically intended for servers that are at / above their intended capacity, but we also use it for servers that haven't received a block for an extended period. The load balancer will route to Healthy servers if any are available, Warning servers only if no healthy servers available, and Unhealthy servers never get routed to.

As our health checks get more nuanced, I think three tiers probably makes sense.

AusIV (commented):
I want to make sure it's widely understood here that this changes our consistency paradigm fairly significantly.

Until now, we've relied on routing sessions consistently to the same provider with the expectation that those providers are going to offer internal consistency.

By doing these block number checks, there's a much higher chance that a provider will get removed from the rotation occasionally just because it won't have heard about a block other providers are reporting yet. That will mean a customers requests will get bounced between different providers a lot more frequently, which should be okay assuming they're actually getting routed to the provider with the latest data available.

But I also want to call out the capacity of individual providers. Right now there's kind of an expectation that any given provider could handle all of the traffic DIN is getting for a network, but if DIN is successful that won't always hold. If we get down to one provider taking all the traffic because it's the only one that has seen the latest block when the healthcheck is done, we could end up overloading a provider. We discussed the complexity of these factors in New York last September, I think @egalano has some notes.

modules/service.go
egalano: Yeah good points from Austin. Lets make sure we're all on the same page on how to merge in these health check improvements without breaking our current consistency guarantees. The intent was to have a deeper health check to guarantee that the node is correctly marked unhealthy based on stalled blocks or lagging updates. We want to keep the same consistency guarantees we've been operating with.

modules/din_upstreams_test.go
natefikru: The logic for the upstream priorities are being tested here. The failover failure was because the provider wasn't being recognized as "unhealthy". as Austin has stated, his updated code should have solved that issue.

modules/service.go
```diff
@@ -0,0 +1,110 @@
...
+		if statusCode > 399 {
+			// if the status code is greater than 399, mark the provider as a failure
+			provider.markPingFailure(s.HCThreshold)
+			continue
+		} else {
+			provider.markPingSuccess(s.HCThreshold)
+		}
+
+		// Consistency health check
+		if s.LatestBlockNumber == 0 || s.LatestBlockNumber < providerBlockNumber {
```
AusIV, line 69: Put some buffer here so that things aren't immediately marked as unhealthy and we can keep a session active.

Don't enter providers back into the rotation until they're caught up to current.

modules/service.go
natefikru: we still don't have this logic in yet, but i added the infrastructure to be able to create this type of non-updating blocknumber check.

modules/service.go
natefikru: i've updated the code with healthy, unhealthy and warning statuses now.

AusIV (approved):

PR #47, opened 2024-09-18
Sync Din Registry Data with Proxy periodically

natefikru:
### Changes made:
- Requires an [updated version of din-go found here](https://github.com/DIN-center/din-sc/pull/44)
- Created registry config to enable registry sync and set other config values.
- Sync registry based on an epoch value
    - Latest block number is retrieved on an interval
    - If the current block number exceeds the epoch floor of the last checked block Epoch Interval, then update the registry
- Introduced a mutex lock on the top level middleware object
    - Safely read and write to the middleware object using the lock
- 2 ways to update network; **If it exists/if it doesn’t exist**  
    - If the network **already exists** in the proxy: 
        - update the top config values (if they are different)
        - loop through the registry providers
            - if the provider exists in the proxy already, then don’t add it
            - if the provider doesn't exist in the proxy already
                - initialize a new provider and use the proxy's specified registry priority level
    - If the network **doesn’t exist** 
        - create a new network with the valid registry values
        - add all the registry providers to the network
            - use the proxy's specified registry priority level
- updated most int values to uint64 for uniformity

### How to Test:

1. Download the DIN-SC project and follow these steps:

Compile and Launch the Local Test Environment:

`pnpm i && pnpm -w build:all && pnpm -w z`

This opens zellij with three panes:
- Left Pane: Runs anvil.
- Upper-Right Pane: Press Enter to deploy the contracts.

2. determine what the din registry contract address is in the top right panel. (should be at the top) and copy it.

3. open project and download dependencies

4. Run the following command ```DIN_REGISTRY_CONTRACT_ADDRESS={contract_address} xcaddy run --config Caddyfile.private --adapter caddyfile```

Caddyfile.private is a local version of the caddyfile that includes private keys. DM nate for a copy of his.


natefikru:
- define the registry node connection details separately from the list of networks/providers in the caddyfile
    - consider replacing the initial registry connection with a proxy version.

go.mod
```diff
@@ -12,20 +15,29 @@ require (
 	github.com/prometheus/client_golang v1.15.1
 	github.com/spruceid/siwe-go v0.2.1
 	github.com/stretchr/testify v1.9.0
-	go.uber.org/zap v1.25.0
+	github.com/zeebo/assert v1.1.0
+	go.uber.org/zap v1.27.0
 )
 
+replace github.com/DIN-center/din-sc/apps/din-go => /Users/natefikru/go/src/DIN-center/din-sc/apps/din-go
```
cds-amal, line 22: This directive should be removed.

natefikru: I'll definitely update the references here once we are able to merge in your + my din-sc code.

cds-amal:
@natefikru, it would be helpful to have a section on how to test this PR.

natefikru:
> @natefikru, it would be helpful to have a section on how to test this PR.

Updated with the instructions

go.mod
natefikru: addressed this. Right now this PR is pointing to the din-sc PR that I have under review right now. I will need to update this to a tagged version of din-sc when that PR is approved.

natefikru:
This is good to merge into master now.

Tests have been written for some reason when i try to use the din-sc version v0.0.2, it resoolves to a hashed version. I'm not sure if that hashed version is the latest version or if it actually is pointing to the v0.0.2 hash.

AusIV (approved):

PR #50, opened 2024-09-30
Remove need for port in caddyfile

AusIV:
The port in the caddyfile is needed by the reverseproxy.Upstream, but causes problems in the healthcheck system with providers that route based on the Host header and don't expect the port number.

This means that provider.host may diverge from provider.upstream.Dial, which we previously assumed would match. I believe I've caught all of the places those were assumed to be the same, but it's something to keep an eye out for during review.


natefikru (approved):

natefikru:
lgtm

PR #52, opened 2024-10-03
remove prometheus health check instantiations

natefikru:
Right now the bill for signoz is exceeding what our expectations are. In an attempt to reduce the amount of metrics being sent to otel/signoz, we are removing our regular health check methods from being required in the DIN proxy.

### DIFF
--- a/modules/service.go
+++ b/modules/service.go
@@ -98,7 +98,6 @@ func (s *service) healthCheck() {
 				// if there is an error getting the latest block number, mark the provider as a failure
 				s.logger.Warn("Error getting latest block number for provider", zap.String("provider", providerName), zap.String("service", s.Name), zap.Error(err), zap.String("machine_id", s.machineID))
 				provider.markPingFailure(s.HCThreshold)
-				s.sendLatestBlockMetric(provider.host, statusCode, provider.healthStatus.String(), providerBlockNumber)
 				return
 			}
 			blockTime = time.Now()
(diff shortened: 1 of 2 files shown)


cds-amal (commented):
This change looks great and clearly helps reduce the amount of data we're sending to OTLE/Signoz -- definitely a step in the right direction! To better understand the impact, could you provide a bit more detail?

What's the current size of the metrics data sent, particularly the portion attributed to the health check/block methods being removed?

Do you have an estimated target for the data size after the changes are implemented?
Even a rough estimate would be helpful, just to set expectations on the reduction. 

This is not a blocker, I just want to know.

natefikru:
> This change looks great and clearly helps reduce the amount of data we're sending to OTLE/Signoz -- definitely a step in the right direction! To better understand the impact, could you provide a bit more detail?
> 
> What's the current size of the metrics data sent, particularly the portion attributed to the health check/block methods being removed?
> 
> Do you have an estimated target for the data size after the changes are implemented? Even a rough estimate would be helpful, just to set expectations on the reduction.
> 
> This is not a blocker, I just want to know.

Thats a good question. I'm just doing the drastic decision of cutting it all off since we aren't really using these metrics right now for anything useful. Currently we are running the health checks[ every 5 seconds for each provider](https://github.com/DIN-center/din-caddy-plugins/blob/9b27752bac1759c1a3e06f722a04be6fdd8f75ec/modules/consts.go#L21)

we have 47 providers and 9 different instances of the proxy running metrics every 5 seconds. That is 5076 metrics per minute, 304560 metrics per hour, and 7309440 metrics per day. we are charged $0.1 per million requests. 

Now that I just did this math, im realizing that this update is insignificant... LOL

PR #56, opened 2024-10-24
Add single default signer instead of provider-specific signers

AusIV:
This lets us configure a single signer and use that anywhere the siwe auth option is used, without having to configure it on a per provider basis. It still lets us define it on a per-provider basis, but if nothing is specified on a provider, the default signer will be used.

The signer is dinMiddleware.DefaultSigner, so it should be available for providers generated from the on-chain registry.

### DIFF
--- a/modules/din_middleware.go
+++ b/modules/din_middleware.go
@@ -67,7 +67,8 @@ func (d *DinMiddleware) Provision(context caddy.Context) error {
 
 	// Initialize the HTTP client for each network and provider
 	httpClient := din_http.NewHTTPClient()
-	for _, network := range d.Networks {
+	for networkname, network := range d.Networks {
+		d.logger.Debug("Registered network", zap.String("name", networkname))
 		network.HTTPClient = httpClient
 		network.logger = d.logger
 		network.PrometheusClient = promClient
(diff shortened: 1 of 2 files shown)


modules/din_middleware.go
```diff
@@ -67,7 +68,8 @@ func (d *DinMiddleware) Provision(context caddy.Context) error {
 
 	// Initialize the HTTP client for each network and provider
 	httpClient := din_http.NewHTTPClient()
-	for _, network := range d.Networks {
+	for networkname, network := range d.Networks {
```
natefikru, line 71: change to camel case `networkName` or just use `name`

modules/din_middleware.go
```diff
@@ -376,6 +416,9 @@ func (d *DinMiddleware) UnmarshalCaddyfile(dispenser *caddyfile.Dispenser) error
 					return dispenser.Errf("expected at least one provider for network %s", networkName)
 				}
 			}
+			if len(d.Networks) == 0 {
```
natefikru, line 419: this isn't necessarily true if the user was trying to only use registry networks.

modules/din_middleware.go
```diff
@@ -231,6 +233,41 @@ func (d *DinMiddleware) UnmarshalCaddyfile(dispenser *caddyfile.Dispenser) error
 	}
 	for dispenser.Next() { // Skip the directive name
 		switch dispenser.Val() {
+		case "siwe-signer":
```
natefikru, line 236: is it possible to see an example of what this would look like in the public`Caddyfile`

AusIV: Done.

modules/din_middleware.go
natefikru: I've explicitly defined this bootstrapped network separately from the caddyfile networks based on a previous conversation that we had.

natefikru (approved):

modules/din_middleware.go
natefikru: maybe in my PR i can make a middleware provision check that makes sure the 

```
len(d.Networks) != 0 && d.registry_enabled == false
```

PR #60, opened 2024-11-25
Feature/vendor din sc

AusIV:
Update the registry integration branch to vendor din-sc, since its repository is currently private.


go.mod
```diff
@@ -2,7 +2,6 @@ module github.com/DIN-center/din-caddy-plugins
 
 go 1.22
 
-toolchain go1.22.8
 
 require (
 	github.com/DIN-center/din-sc/apps/din-go v0.0.0-20241029204756-63334786749d
```
natefikru, line 7: what build version are you bringing in from din-sc, i can't tell. is it just latest? i see it as `v0.0.0-20241029204756-63334786749d`

natefikru: im currently pondering if there is a way to do this automatically instead of having to statically bring the files over each time.

AusIV:
So, it looks like vendoring is all or nothing within go. It didn't turn up in my test builds through caddy because Caddy treats the plugins as a package rather than the build target, so it's not concerned about the plugins' go.mod

In any case, I just renamed the "vendor" directory to "upstream" to get the tests to pass. The absence of a "vendor" directory made Go not worry about not being able to find other packages in the go.mod.

go.mod
AusIV: Based on the conversation with Tom and Amal, din-sc should be public the first week in january, so we won't have to worry about this for too long.

PR #61, opened 2024-11-27
Feature: ensure block numbers do not drastically change for greater or lesser

natefikru:
Previously, we have seen bugs amongst providers where for example if the general block range is around 100-105, a provider may fail and jump to 500, then back down to 103. 

As a result, this update implements the concept of a `blockNumberDelta` value that makes it so providers can't go over or under and a certain range when it comes to latest block numbers that they are reporting.

For example, now if the `blockNumberDelta = 10`, if a provider jumps up to 500, then they are deemed as `unhealthy` and need to go through the healthiness process to prove that they are healthy again.

I also did a general refactor of the healthCheck logic by abstracting out the checks into their own functions and reflecting this in the tests themselves.

### DIFF
--- a/modules/consts.go
+++ b/modules/consts.go
@@ -23,6 +23,7 @@ const (
 	DefaultHCThreshold             = 2
 	DefaultHCInterval              = 5
 	DefaultBlockLagLimit           = int64(5)
+	DefaultBlockNumberDelta        = int64(10)
 	DefaultMaxRequestPayloadSizeKB = int64(4096)
 	DefaultRequestAttemptCount     = 5
 
--- a/.github/workflows/e2e_integration_test/Caddyfile
+++ b/.github/workflows/e2e_integration_test/Caddyfile
@@ -16,7 +16,7 @@
 						https://ethereum-rpc.publicnode.com:443 {
 							priority 0
 						}
-						https://eth.rpc.blxrbdn.com:443 {
+						https://api.zan.top/eth-mainnet {
 							priority 0
 						}
 					}
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,6 @@
 module github.com/DIN-center/din-caddy-plugins
 
-go 1.22
-
+go 1.21.7
 
 require (
 	github.com/DIN-center/din-sc/apps/din-go v0.0.0-20241029204756-63334786749d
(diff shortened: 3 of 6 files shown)


modules/consts.go
```diff
@@ -23,6 +23,7 @@ const (
 	DefaultHCThreshold             = 2
 	DefaultHCInterval              = 5
 	DefaultBlockLagLimit           = int64(5)
+	DefaultBlockNumberDelta        = int64(10)
```
natefikru, line 26: This value is definitely subject to change, let me know if this makes sense or not.

AusIV:
First, something's weird about this PR. It's reporting that there are 60 commits being merged. It looks like it's trying to merge in every commit since you started working on the registry implementation, but that's already been merged so I'm not sure what's going on there.

In terms of the content of this PR, there's a scenario here that a proxy could never recover from:

Suppose we had a network with just one provider. That provider is having syncing issues, and has fallen considerably behind the network. The provider resolves their syncing issues and catches up quickly (maybe by replacing the out-of-sync server with an in-sync one, maybe by just processing blocks quickly). Now we have a leap forward. On a network that has blocks every two seconds, if the provider was down for an hour, it could leap forward 1800 blocks quite abruptly.

Under the current implementation, if latestBlock is 1,000,000 and the sole provider suddenly leaps forward to 1,001,800, I believe we'll always reject the leap forward, but because it was a correct leap forward it will never go back into the expected range, and thus will never resolve.

I've also seen consensus issues with networks that caused issues like this on large scales. For example if a hard fork happens and people haven't upgraded their nodes, or some weird consensus issue happens that effects some clients but not others, it could impact multiple providers at the same time, and as they correct the issue they'll jump forward abruptly. If all of the providers were effected by the same issue, we'd probably see them all get stuck at the same place, then leap forward one provider at a time as they correct the issue. But if latestBlock never gets updated because the leap was too big, all the providers could be caught up again and still marked as unhealthy.

I think rather than comparing it to the latestBlock, we should consider comparing it to the median block (or 75th percentile block, or some other perecentile) across all the providers. If we only have one provider, the leap forward would be considered correct. If we have two providers and one leaps forward, the one that's way ahead is unhealthy. If we have two providers and one leaps forward then the other follows, they'd be considered healthy once they had both caught up. In the "network having consensus issues" scenario above, this would mean that the providers who had corrected the issue wouldn't be considered healthy until more providers have corrected the issue, but once enough providers correct the issue the proxy would resolve to healthy.

I think a `DefaultBlockNumberDelta` value of 10 is fine if we're comparing to an aggregate of the existing providers, but I think we'd need that number to be in the tens of thousands (if not hundreds of thousands) if we're going to compare to latestBlock. (This would still usually catch the odd scenario of "Provider X just started serving a different network")

natefikru:
regarding the commits, i merged it in with the registry code, so these are all of those commits. I'll just make sure to squash this PR when i merge it

AusIV (approved):

PR #67, opened 2024-12-17
Smart Routing: add flag for evm speed, keep track of latest and earliest block number

natefikru:
This commit improves the monitoring capabilities of EVM endpoints while maintaining 
backward compatibility with existing functionality.

Key Changes:
1. Added EVMSpeed Feature flag Support
   - Introduced EVMSpeedEnabled flag for providers
   - Added DefaultEVMSpeedEnabled constant (default: false)
   - Added DefaultGetBlockNumberMethod constant

2. Enhanced Block Number Tracking
   - Added capability to track both earliest and latest block numbers
   - Implemented binary search algorithm to find earliest block
   - only check earliest blockNumber on initialization; binary search doesn't need to be ran on each health check.
   - earliestBlockNumber check starts at 1 so as to not confuse the genesis block number value of `0` with the default `0`value of the `provider.earliestBlockNumber` struct.
   - Added block number storage functionality for providers

3. Provider Enhancements
   - Moved block number fetching logic from network to provider
   - Added methods for saving earliest and latest block numbers
   - Enhanced error handling for block number queries

4. Testing Improvements
   - Added comprehensive tests for hex/int conversions
   - Added tests for block number tracking
   - Added tests for EVMSpeed functionality
   - Enhanced provider and network test coverage

5. Configuration Updates
   - Added evm_speed_enabled option to Caddyfile parser
   - Updated provider struct with new fields

### DIFF
--- a/modules/network.go
+++ b/modules/network.go
@@ -109,16 +109,20 @@ func (n *network) healthCheck() {
 			}
 
 			if n.EVMSpeedEnabled {
-				earliestBlockNumber, statusCode, err := provider.getEarliestBlockNumber(n.HCMethod, n.RequestAttemptCount)
-				if err != nil {
-					n.handleBlockNumberError(providerName, provider, statusCode, providerBlockNumber, err)
-					return
-				}
+				// Only check and set earliest block number if it hasn't been set yet.
+				// earliest block to check will be block 1.
+				if provider.earliestBlockNumber != 0 {
+					earliestBlockNumber, statusCode, err := provider.getEarliestBlockNumber(n.HCMethod, n.RequestAttemptCount)
+					if err != nil {
+						n.handleBlockNumberError(providerName, provider, statusCode, providerBlockNumber, err)
+						return
+					}
 
-				err = provider.saveEarliestBlockNumber(earliestBlockNumber)
-				if err != nil {
-					n.logger.Error("Error saving earliest block number", zap.String("provider", providerName), zap.String("network", n.Name), zap.Error(err), zap.String("machine_id", n.machineID))
-					return
+					err = provider.saveEarliestBlockNumber(earliestBlockNumber)
+					if err != nil {
+						n.logger.Error("Error saving earliest block number", zap.String("provider", providerName), zap.String("network", n.Name), zap.Error(err), zap.String("machine_id", n.machineID))
+						return
+					}
 				}
 			}
 
--- a/modules/network.go
+++ b/modules/network.go
@@ -111,8 +111,8 @@ func (n *network) healthCheck() {
 			if n.EVMSpeedEnabled {
 				// Only check and set earliest block number if it hasn't been set yet.
 				// earliest block to check will be block 1.
-				if provider.earliestBlockNumber != 0 {
-					earliestBlockNumber, statusCode, err := provider.getEarliestBlockNumber(n.HCMethod, n.RequestAttemptCount)
+				if provider.earliestBlockNumber == 0 {
+					earliestBlockNumber, statusCode, err := provider.getEarliestBlockNumber(DefaultGetBlockNumberMethod, n.RequestAttemptCount)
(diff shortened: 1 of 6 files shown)


AusIV:
Are we actually doing anything with the earliest block number once we have it? And aside from enabling the collection of earliest block number, what does enabling EVM Speed do?

natefikru:
> Are we actually doing anything with the earliest block number once we have it? And aside from enabling the collection of earliest block number, what does enabling EVM Speed do?

EVM speed is the Infura service that utilizes cacheing to expose many different services to us including a resolver service that will get the correct block number based on a hash or other values etc. 

The tracking of the earliest/latest will help us utilize the evm-speed service to route traffic to the proper providers that support the specific block by #

modules/network.go
```diff
@@ -96,14 +94,38 @@ func (n *network) healthCheck() {
...
+			}
+
+			if n.EVMSpeedEnabled {
+				// Only check and set earliest block number if it hasn't been set yet.
+				// earliest block to check will be block 1.
+				if provider.earliestBlockNumber == 0 {
+					earliestBlockNumber, statusCode, err := provider.getEarliestBlockNumber(DefaultGetBlockNumberMethod, n.RequestAttemptCount)
+					if err != nil {
+						n.handleBlockNumberError(providerName, provider, statusCode, providerBlockNumber, err)
+						return
```
ggarri, line 118: If I am understand the code correctly, implementing an early-exit strategy for this opt-in mode for evm-speed might cause underline health check to never complete, does this might case providers to be considered as unhealthy? if so, I would not perform an early exit but just a WARN/ERR logging to make this operation failure visible but no becoming a "core" check of the health validation.

modules/provider.go
```diff
@@ -128,3 +135,205 @@ func (p *provider) Warning() bool {
...
+	if err != nil {
+		return 0, 0, err
+	}
+	return blockNumber, *statusCode, nil
+}
+
+// getEarliestBlockNumber gets the earliest block number from the provider
+// First checks block 1, then uses binary search if not found
+// This is only enabled for EVM based networks
+func (p *provider) getEarliestBlockNumber(getBlockNumberMethod string, retryCount int) (int64, int, error) {
```
ggarri, line 219: I cannot find any usage of `retryCount`

ggarri, line 219: I think you can simplify this by doing the binary search as part of the main function and rewriting it as recursive.

```suggestion
func (p *provider) getEarliestBlockNumber(getBlockNumberMethod string, lastFound, left, right int) (int64, int, error) {

if left >= right {
   return lastFound
}

mid := left + (right-left)/2

// Create the JSON-RPC payload for getting the block
payload := []byte(fmt.Sprintf(`{"jsonrpc":"2.0","method":"%s","params":["%s", false],"id":1}`,
    getBlockNumberMethod, intToHex(mid)))

....
			
if result, ok := respObject["result"]; ok && result != nil {
return    getEarliestBlockNumber(getBlockNumberMethod string, mid, left, mid - 1)
} else {
return    getEarliestBlockNumber(getBlockNumberMethod string, mid, mid +1, left)
}
....


```

modules/provider.go
```diff
@@ -128,3 +135,205 @@ func (p *provider) Warning() bool {
...
+	return p.binarySearchEarliestBlock(getBlockNumberMethod)
+}
+
+func (p *provider) saveLatestBlockNumber(blockNumber int64) error {
+	p.latestBlockNumber = uint64(blockNumber)
+	return nil
+}
+
+func (p *provider) saveEarliestBlockNumber(blockNumber int64) error {
+	p.earliestBlockNumber = uint64(blockNumber)
```
ggarri, line 266: This operation (same for `saveLatestBlockNumber()` are not racing safe as it can cause concurrency read/write. I would recommend to use mutex to prevent that

Issue #68, opened 2024-12-18
Dynamic Load Balancing with Watcher Scores


noleto:
### Current Situation:
The DIN Router currently routes traffic evenly among providers of the same priority level. “Evenly” means each eligible and healthy provider gets an equal share of the traffic. There’s no consideration of provider performance or data quality in this distribution.

### Desired Enhancement:
Introduce a dynamic load balancing policy that reflects how well each provider is performing in terms of data consistency and latency over time  so better-performing providers receive a proportionally larger share of requests. The ultimate goal: improve service quality for DIN developers.

### Introducing Watcher Score

Providers will be scored based on a set of criteria, forming the "Provider Watcher Score." This score is calculated using data provided by Watchers, which perform regular checks and latency measurement. The criteria currently monitored by Watchers include:

* Block Number Consistency (BNC): Ensuring that the block number either monotonically increases or remains consistent.
* Consistency of Non-State Data for a Block (CNSB): Verifying that non-state data for a given block hash remains consistent over time.
* Request Latency (RLAT): Measuring the elapsed time (in milliseconds) between sending a request and receiving a response, targeting the provider endpoint directly (without the DIN Router).

These metrics are evaluated per network and per provider. The formula to compute the Provider Watcher Score for a provider (e.g., P1) in a given network is:

$Φ({P1}) = BNC(P1) ⋅ W_{bnc} + CNSB({P1}) ⋅ W_{cnsb} + L({P1} | {P(95)})⋅  W_{rlat}$,

where:

* Φ (Phi) is the Watcher Score (range: [0,1])
* $W_{bnc}$, $W_{cnsb}$, $W_{rlat}$ are the weights assigned to each criterion (weights sum to 1)
* BNC, CNSB, and RLAT are normalized scores (range: [0,1]) for Block Number Consistency, Consistency of Non-State Data, and Latency at the 95th percentile, respectively.

###  Integration with DIN Router:
The DIN Router will:
* Maintain a table mapping each provider and network to a computed Watcher Score.
* On receiving a request for network N, if no session affinity applies, it will consider all available provider that may satisfy the request. 
* Based on the watcher score, the router will produce a set of provider weights that will determine the traffic allocated for each provider. For example: If $Φ(P1)=0.9$, $Φ(P2)=0.8$, $Φ(P3)=0.55$, the router will produce a set of traffic weights: $TW(P1)=0.4$, $TW(P2)=0.35$, $TW(P1)=0.25$.
[...]

PR #71, opened 2025-01-07
[Feature] Enable Reputation-Driven Adaptive Load Balancing

noleto:
This PR brings the dynamic routing feature to DIN Router as detailed in ticket #68. It is strongly recommended to read[ the design doc](https://github.com/DIN-center/din-caddy-plugins/blob/feature/68-reputation-driven-adaptive-load-balancing/docs/dip-001/dip-001-smart-routing.md) before reviewing so that terminology and the overall workflow will be clear.

The PR is mainly reviewed in two parts, and following this order will make reviewing easier for you:
- Implementation of the reputation score module (in `lib/reputationscore`)
- Then, the smart routing workflow that leverages the reputation score (in `modules`)

The core logic was extensively tested using unit tests to ensure reproducibility (as we are using some randomness in the provider selection). See coverage report.

Manual integration tests are also possible based on the `Caddyfile.private` file that is not in the repo. You can ask me for it. This can be achieved by:
- Cloning the repo (or updating the code to this branch feature)
- Starting a local Caddy instance or preferably using this Docker command: `docker run --rm -p 8000:8000 -v ./Caddyfile.private:/etc/caddy/Caddyfile din-caddy caddy run --config /etc/caddy/Caddyfile --adapter caddyfile`
- Triggering a few queries:

```shell
curl --location 'localhost:8000/bsc-mainnet' \
--header 'Content-Type: application/json' \
--data '{"method":"eth_blockNumber","params":[],"id":1,"jsonrpc":"2.0"}'
```

Loggers are in Debug mode so you can follow the workflow. Loggers related to this feature start with:
- `[REPUTATION_SCORE]...` for all loggers related to reputation score logic
- `[SMART_ROUTING]...` for all loggers related to the DIN proxy logic that uses provider scores to dynamically load balance the traffic


modules/watcherapi_client.go
```diff
@@ -0,0 +1,336 @@
...
+
+	"github.com/go-playground/validator/v10"
+)
+
+// CheckParams represents the parameters for a check request
+type CheckParams struct {
+	CheckID    string   `query:"check_id" validate:"required,oneof=blockNumberConsistency blockNonStateConsistency"`
+	Network    string   `query:"network" validate:"required"`
+	ProviderID []string `query:"provider_id"`
+	Region     []string `query:"region" validate:"dive,oneof=NorthAmerica Latam Europe MiddleEastAfrica AsiaPacific"`
```
natefikru: havent used the validator lib yet, this is an interesting addition.

noleto:
Reputation Score and Traffic Distribution Separation

After further analysis and simulation runs, it is evident that the final score used for traffic distribution should be dissociated from the Reputation Score engine. The Reputation Score engine should remain agnostic to the application of the score. Instead, the transformation from the Reputation Score to traffic distribution weights should be handled separately.

Additionally, the percentage share of total traffic is not inherently a score but a derived transformation from the Reputation Score into traffic weights. This ensures that the methodology used for scoring remains independent of how it is ultimately applied.

Another crucial aspect is that some providers may be monitored by the Watcher but may not currently be served by the DIN Router. If the final traffic weight is directly derived from the Reputation Score as it is today, it would result in an incomplete picture. Therefore, a distinct layer must handle the conversion of Reputation Scores into traffic weights while ensuring consistency across all active and monitored providers.

ping @natefikru

modules/din_middleware_helpers.go
```diff
@@ -304,6 +323,27 @@ func (d *DinMiddleware) updateNetworkData(network *network) {
 	}
 }
 
+func (d *DinMiddleware) GetOrCreateWatcherClient() watcher.IWatcherAPIClient {
```
natefikru, line 326: This function should Probably be a public function in the watcher library

Caddyfile
```diff
@@ -329,6 +329,13 @@
 			}
 			din_registry {
 				registry_enabled false
+				smart_routing {
```
natefikru, line 332: we shouldn't use the term smart routing to describe this. The smart routing namespace is reserved for the concept of routing to different providers based on block depth. We should use something more relevant here like `weightedRouting` `reputationScoring` or  `smartScoring`

lib/reputationscore/types.go
```diff
@@ -0,0 +1,137 @@
...
+
+// Equal returns true if the two ProviderMetrics are equal
+func (m *ProviderMetric) Equal(other *ProviderMetric) bool {
+	if m == nil || other == nil {
+		return m == other
+	}
+	return m.metricID == other.metricID &&
+		m.providerName == other.providerName &&
+		m.providerURL.String() == other.providerURL.String() &&
+		math.Abs(m.value-other.value) < 0.00001 &&
```
natefikru: how do you come to the value of 0.0001? Can this value potentially change? if so it should proabably be a constant

lib/reputationscore/combiners.go
```diff
@@ -0,0 +1,33 @@
...
+	// Find the most recent metric timestamp
+	maxLastUpdated := time.Time{}
+	for _, metric := range metrics {
+		if metric.LastUpdated().After(maxLastUpdated) {
+			maxLastUpdated = metric.LastUpdated()
+		}
+	}
+	finalScore, err := NewScore(weightedValue, maxLastUpdated)
+
+	return finalScore, err
```
natefikru: can just write this as 
`return NewScore(weightedValue, maxLastUpdated)`

lib/reputationscore/combiners.go
```diff
@@ -0,0 +1,33 @@
...
+	"time"
+)
+
+// A weighted combiner is a base implementation of combiner that uses weights to aggregate metrics.
+type WeightedCombiner struct {
+	Weights map[string]float64
+}
+
+func (c *WeightedCombiner) CombineMetrics(metrics []*ProviderMetric) (*Score, error) {
+	weightedValue := 0.0
```
natefikru: we could also add a total weight validation step as well:

```
	// Weight validation
	totalWeight := 0.0
	for _, weight := range c.Weights {
		totalWeight += weight
	}
	if totalWeight < 0.99 || totalWeight > 1.01 {
		return NewEmptyScore(), fmt.Errorf("weights must sum to 1.0, got %f", totalWeight)
	}
	```

File: lib/reputationscore/transformers.go [hunk 7]
```diff
@@ -0,0 +1,139 @@
...
+)
+
+// ShareOfTotalTransformer adjust the input score so the sum of all scores is 1.
+// In other words, adjusted score represents the percentage of the total score.
+type ShareOfTotalTransformer struct {
+}
+
+func (t *ShareOfTotalTransformer) TransformScore(scores map[string]*Score) (map[string]*Score, error) {
+	if len(scores) == 0 {
+		return scores, nil
```
natefikru: we shouldn't return an error here?

lib/reputationscore/transformers.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	// Update the previous scores with the smoothed scores
+	t.previousScores = smoothedScores
+
+	return smoothedScores, nil
+}
+
+func NewEWMATransformer(alpha float64) *EWMATransformer {
+	return &EWMATransformer{alpha: alpha}
```
natefikru: we could add some validation here:

```
if alpha < 0 || alpha > 1 {
    // error
}

File: lib/reputationscore/reputationscore_manager.go [hunk 9]
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	rm := NewEmpty(logger)
+	for network := range uniqueNetworksMap {
+		rm.AddNetworkWithBuiltInFormula(network, client)
+	}
+	return rm
+}
+
+func (rm *ReputationScoreManager) ComputeScores() error {
```
natefikru: depending on how much data is being computed, if it is causing too much memory usage, we could potentially batch the computation in chunks to ease some of the memory management pressure. Probably not something that we need right now but it could be useful in the future for when we scale up our operations.

natefikru (commented):
Overall great job at implementing the core logic, this looks great!

Caddyfile
noleto: okay, didn't know this was reserved for another feature. probably then the best short names for this feature are: `adaptive_lb` or `score_based_lb`

lib/reputationscore/types.go
noleto: this is a common strategy to compare floats (add an arbitrary `tolerance` based on the domain) => https://golangbyexample.com/comparing-floating-point-numbers-go/

lib/reputationscore/transformers.go
noleto: by definition, transforming an empty set of scores returns the same empty, but totally arbitrary decision made by the developer (myself :D)

lib/reputationscore/reputationscore_manager.go
noleto: I've tested with all current live providers, and memory pressure is very low since we make sequential calls and do not store any cache. It was designed to be used in a "relaxed" context. By "relaxed," I mean it is more CPU-bound than memory-bound. Definitely something that is not tailored for real-time needs.

lib/reputationscore/combiners.go
noleto: Done

lib/reputationscore/combiners.go
noleto: Agree, added more robust checks related to weights (not only checking sum is 1.0 but also checking the weights size matches the size of metrics.

modules/din_middleware_helpers.go
noleto: The goal of this function is to handle a cache at the DinMiddleware level for WatcherClient, hence its name "GetOrCreateWatcherClient"

noleto:
closed this PR because working in a new version https://github.com/DIN-center/din-caddy-plugins/pull/150 (also the original branch was deleted for security reasons)

PR #77, opened 2025-01-28
Robust Health Checks

natefikru:


modules/network.go
```diff
@@ -329,37 +346,64 @@ func (n *network) getLatestBlockNumber(httpUrl string, headers map[string]string
 	return blockNumber, *statusCode, nil
 }
 
-func (n *network) close() {
-	close(n.quit)
-}
+func (n *network) getChainID(httpUrl string, headers map[string]string, ac auth.IAuthClient) (int64, int, error) {
```
AusIV: Several things here: 

First, Geth encodes the chainid as a big.Int, implying it can exceed int64. The idea is kinda horrifying, but especially with app chains I could see this becoming a factor.

Second, Bitcoin and Solana don't have numeric Chain IDs. Bitcoin has a string chain name buried in the object returned by its [getblockchaininfo](https://developer.bitcoin.org/reference/rpc/getblockchaininfo.html) method. Solana has a [getGenesisHash](https://solana.com/docs/rpc/http/getgenesishash) method that I should uniquely identify the chain (barring chainsplits :crossed_fingers:).

I think we should probably change the return value to a string, not worry about converting it to an integer (we don't need to do math with it or make sure it's monotonically increasing like with block numbers). Then maybe add a "keys" parameter where we could do something like:

```
// Given var response map[string]interface{}
value := response
for _, key := range keys {
   if v, ok := value[key]; ok {
      switch tv := tv.(type) {
         case map[string]interface{}:
             value = tv
         case string:
            return tv
         default:
            // Error unknown type
      }
   } else {
     //error key missing
   }
}
```

Which would let us find some unique identifier for each chain.

modules/din_middleware.go
```diff
@@ -494,6 +494,22 @@ func (d *DinMiddleware) UnmarshalCaddyfile(dispenser *caddyfile.Dispenser) error
...
+					case "chainid_method":
+						dispenser.Next()
+						d.Networks[networkName].ChainIDMethod = dispenser.Val()
+					case "chain_id":
+						dispenser.Next()
+						chainID, err := strconv.ParseInt(dispenser.Val(), 10, 64)
+						if err != nil {
+							return fmt.Errorf("invalid expected chain ID: %v", err)
+						}
+						// Skip chain ID validation for Solana networks
```
AusIV: Do you have a way to get a chain id for bitcoin?

modules/network.go
```diff
@@ -81,144 +84,209 @@ func (n *network) startHealthcheck() {
...
+	}
+
+	// Check chain ID - this is a critical check that should always result in Unhealthy
+	if !n.verifyChainID(provider) {
+		n.logProviderWarning("Provider has incorrect chain ID", provider)
+		return Unhealthy
+	}
+
+	// if provider has no block history, set it to healthy
+	if len(provider.BlockHistory()) == 0 {
```
AusIV, line 162: Do we want to do this? It might not hurt to require some amount of block history to verify that a provider is doing what it's supposed to before calling it healthy.

modules/network.go
```diff
@@ -81,144 +84,209 @@ func (n *network) startHealthcheck() {
...
+// Falls back to warning providers if no healthy providers are available
+func (n *network) getLatestHealthyBlock() int64 {
+	var latestBlock int64
+	hasHealthyProviders := false
+
+	// First try to get block from healthy providers
+	for _, provider := range n.Providers {
+		if provider.healthStatus == Healthy && len(provider.BlockHistory()) > 0 {
+			history := provider.BlockHistory()
+			lastBlock := history[len(history)-1].blockNumber
```
AusIV: Does history get added to when the number goes down? If yes, we might want to take the max block number from this list, rather than the last one in the list

modules/network.go
```diff
@@ -81,144 +84,209 @@ func (n *network) startHealthcheck() {
...
+		}
+	}
+
+	// Check for stalling - all blocks in history are identical
+	if n.isStalled(provider) {
+		if n.allProvidersStalled() {
+			// This signifies a network outage
+			n.logProviderWarning("All providers are stalled", provider)
+			if Warning > worstStatus {
+				worstStatus = Warning
```
AusIV, line 197: I don't think we should be setting all providers to warning because all providers are stalled.

Being significantly lagged but making progress puts you in a warning status. If the network is stalled, but one provider is way behind the others, having them all in warning would mean the provider that's way behind is just as eligible for traffic as the ones that are in the best available state.

I'm still generally of the opinion that isStalled should only be checked in the context of a server that is also lagged, and stalled + lagged should lead to unhealthy, but being stalled without being lagged says nothing about a provider.

modules/din_middleware.go
natefikru: it doesn't look like it!

modules/network.go
natefikru: Thats valid.

modules/network.go
natefikru: good point!

modules/network.go
natefikru: I'll update that.

modules/consts.go
```diff
@@ -20,12 +20,14 @@ const (
...
 	DefaultHCMethod                = "eth_blockNumber"
+	DefaultChainIDMethod           = "eth_chainId"
 	DefaultHCThreshold             = 2
 	DefaultHCInterval              = 5
 	DefaultBlockLagLimit           = int64(5)
-	DefaultBlockNumberDelta        = int64(10)
+	DefaultBlockJumpLimit          = int64(100)
 	DefaultMaxRequestPayloadSizeKB = int64(4096)
 	DefaultRequestAttemptCount     = 5
+	DefaultBlockHistorySize        = 10
```
natefikru, line 30: Need to add support for this value in the caddy file.

modules/network.go
```diff
@@ -81,144 +84,209 @@ func (n *network) startHealthcheck() {
...
-			}
-
-			if n.blockNumberDeltaHealthCheck(providerName, provider, providerBlockNumber) {
-				return
-			}
+	// First get latest blocks from all providers and update their history
+	for _, provider := range n.Providers {
+		blockNum, statusCode, err := n.getLatestBlockNumber(provider.HttpUrl, provider.Headers, provider.AuthClient())
+		if err != nil {
+			n.logger.Warn("Error getting latest block number for provider",
```
natefikru: call the new log warning wrapper function

modules/consts.go
```diff
@@ -20,12 +20,14 @@ const (
 
 	// Health check constants
 	DefaultHCMethod                = "eth_blockNumber"
+	DefaultChainIDMethod           = "eth_chainId"
```
ggarri, line 24: minor: instead of default you could have called `evmChainIDMethod` as all evm network will have the same method name

modules/din_middleware.go
ggarri: for the sake of validating we are serving the right solana network  we could use a different strategy by validating the genesisHash https://solana.com/docs/rpc/http/getgenesishash

modules/din_middleware_helpers.go
```diff
@@ -211,6 +212,20 @@ func (d *DinMiddleware) syncNetworkConfig(regNetwork *din.Network, network *netw
...
 	}
 
+	// Sync chain ID if provided in registry
+	registryChainID := regNetwork.NetworkConfig.ChainID
+	if !strings.Contains(network.Name, "solana") {
+		if registryChainID == 0 {
+			return nil, fmt.Errorf("chain ID is required in registry for network %s", network.Name)
+		}
+		// Verify chain ID matches if already set
+		if network.ChainID != 0 && network.ChainID != registryChainID {
```
ggarri: under which cases you expected `network.ChainID != 0 ` as I see, that should never be the case.

modules/din_middleware_helpers.go
```diff
@@ -211,6 +212,20 @@ func (d *DinMiddleware) syncNetworkConfig(regNetwork *din.Network, network *netw
...
+			return nil, fmt.Errorf("registry chain ID (%d) does not match configured chain ID (%d) for network %s",
+				registryChainID, network.ChainID, network.Name)
+		}
+		network.ChainID = registryChainID
```
ggarri, line 224: why updating the network.chainID witht the value in the config?  does that not fake the results?

modules/network.go
```diff
@@ -81,144 +84,209 @@ func (n *network) startHealthcheck() {
...
-				return
-			}
-
-			if n.blockNumberDeltaHealthCheck(providerName, provider, providerBlockNumber) {
-				return
-			}
+	// First get latest blocks from all providers and update their history
+	for _, provider := range n.Providers {
+		blockNum, statusCode, err := n.getLatestBlockNumber(provider.HttpUrl, provider.Headers, provider.AuthClient())
+		if err != nil {
```
ggarri, line 94: As I see it, getting a single error from the provider should not lead to an Unhealthy status right away, as it might be an occasional state when this happens, and updating the provider status to unhealthy might have larger implications for the service reliability. In addition I see don't distinguish between HTTP errors and payload decoding errors, and we should do to prevent us making mistakes in our assumptions on how to unmashall a retrieved payload, for instance in non-evm networks like Solana.

IMO, firstly, we should implement a fairly simple retry to fetch the latest block in case of 5XX HTTP errors, so we don't get misled by a temporal service errors. Secondly, if a provider was healthy I would rather go firstly to Warning and then to Unhealthy, so that we have a more gradual state change. This last idea, I would rather suggest it to apply to all cases where we change from Healthy to Unhealthy

modules/network.go
```diff
@@ -81,144 +84,209 @@ func (n *network) startHealthcheck() {
...
-	if referenceBlock > n.latestBlockNumber {
-		n.latestBlockNumber = referenceBlock
+// getLatestHealthyBlock returns the highest block number among healthy providers
+// Falls back to warning providers if no healthy providers are available
+func (n *network) getLatestHealthyBlock() int64 {
+	var latestBlock int64
+	hasHealthyProviders := false
+
+	// First try to get block from healthy providers
+	for _, provider := range n.Providers {
```
ggarri, line 250: To save one iteration over all providers, why no having different variable for latest depending of the node status `latestBlockFromHealthy`, `latestBlockForWarning` and ` `latestBlockForUnhealthy` and base on that you can first return "latestBlockFromHealthy" if different than 0, if not "latestBlockForWarning", and at last "latestBlockForUnhealthy" if all remaining at 0. That way you guarantee you already have a reference value for latest

modules/network.go
```diff
@@ -81,144 +84,209 @@ func (n *network) startHealthcheck() {
...
+				zap.Int64("provider_block", currentBlock),
+				zap.Int64("network_block", latestNetworkBlock))
+			if Warning > worstStatus {
+				worstStatus = Warning
+			}
+		}
+
+		// Check if block is too far ahead (block jump)
+		blockJump := currentBlock - int64(latestNetworkBlock)
+		if blockJump > n.BlockJumpLimit {
```
ggarri, line 182: I am still very concerned on this scenario, having a block jump can be a scenario we hit due to temporal network outages. If we have a block jump, and the chainID of the provider matches the expected value, I would rather skip the state change to Unhealthy.

Alternatively, we could move the provider to Warning in this case, so that if all providers produces the same jump behavior, since all of them end up as Warning state, we wont have an scenario where all providers end up flagged as Unhealthy at once causing potential service interruption

modules/network.go
ggarri: I also believe this check can be skipped entirely. The `blockLag` check should be sufficient to degrade the status of a provider to Warning/Unhealthy if other providers are moving forward on the latest block but in case all providers are stack at the same block it likely implicates a global network issue and in that case we don't want to be flagged all providers as unhealthy.

modules/network.go
```diff
@@ -81,144 +84,209 @@ func (n *network) startHealthcheck() {
...
-			n.logger.Warn("Provider returned an error status code", zap.String("provider", providerName), zap.String("network", n.Name), zap.Int("status_code", statusCode), zap.String("machine_id", n.machineID))
-			provider.markPingFailure(n.HCThreshold)
+			// This signifies a provider outage
+			n.logProviderWarning("Provider is stalled while others are progressing", provider)
+			return Unhealthy // Stalling when others aren't is always Unhealthy
+		}
+	}
+
+	// Check monotonicity as quality indicator
+	if !n.isMonotonic(provider) {
```
ggarri: As I mentioned before, breaking monoticity rule should not implicate a degradation to Unhealthy, as this scenario can happen often, even with long interval checks if the provider runs multiple nodes and one of, for whatever reason has temporal peering issues. This is an scenario we experimented with out backing nodes at Infura for instance.

My suggestion would be to "penalize" this behaviour to WARNING status, but not to UNHEALTHY

modules/network.go
```diff
@@ -329,37 +346,64 @@ func (n *network) getLatestBlockNumber(httpUrl string, headers map[string]string
 	return blockNumber, *statusCode, nil
 }
 
-func (n *network) close() {
-	close(n.quit)
-}
+func (n *network) getChainID(httpUrl string, headers map[string]string, ac auth.IAuthClient) (int64, int, error) {
+	payload := []byte(fmt.Sprintf(`{"jsonrpc":"2.0","method": "%s","params":[],"id":1}`, n.ChainIDMethod))
```
ggarri, line 361: we should probably replace this evm-like payload for something more generic which is passed as argument to make sure we can cover cases like Solana and Bitcoin. 

In addition, we could rename the method from `getChainID` to `getUniqueID` so that we don't make assumptions about the method we use to check if the provider is serving the expected network data

modules/din_middleware_helpers.go
natefikru: This would be if solana initiated as a network, or any higher level service. This could potentially be solved with some sort of boolean, and not an explicit call

modules/din_middleware_helpers.go
natefikru: Where else would the initial source of truth chain I'd be defined. There needs to be a canonical reference of what is the correct chain id

modules/network.go
natefikru: updated to store chainID as a string and am supporting those methods.

modules/din_middleware.go
natefikru: for bitcoin im using the chain value in the object here [getblockchaininfo](https://developer.bitcoin.org/reference/rpc/getblockchaininfo.html) 

its not perfect, but i think its better than nothing.

modules/network.go
natefikru: updated

modules/network.go
natefikru: im still a little confused about the keys parameter that you. mentioned.

modules/consts.go
natefikru: i think it makes sense to keep this as is just to maintain the pattern that is set up already.

modules/din_middleware_helpers.go
natefikru: removed explicit call

modules/din_middleware_helpers.go
natefikru: i guess i see your point, this value should actually technically never be updated. This assumes that we (DIN) input the data perfectly on the registry the first time. I'm essentially building for potential failure here if we ever have to rollback network data and have routers that are currently running update the value in flight.

modules/network.go
natefikru: implemented the retrys.

Regarding going to warning vs unhealthy, This is a din product level decision, I'll bring this up with our team to discuss. Good feedback though, thank you!

modules/network.go
natefikru: addressed.

modules/network.go
natefikru: Ill take this into account and bring to the team for discussion.

modules/network.go
natefikru: I updated it so that the code checks if a provider's blockchain has stalled (stopped producing new blocks) and handles it differently based on whether all providers are stalled (indicating a network-wide issue, marked as Warning) or just this specific provider is stalled while others continue (indicating a provider-specific issue, marked as Unhealthy).

```

	// Check for stalling - all blocks in history are identical
	if n.isStalled(provider) {
		if n.allProvidersStalled() {
			// This signifies a network outage
			n.logProviderWarning("All providers are stalled", provider)
			if Warning > worstStatus {
				worstStatus = Warning
			}
		} else {
			// This signifies a provider outage
			n.logProviderWarning("Provider is stalled while others are progressing", provider)
			return Unhealthy // Stalling when others aren't is always Unhealthy
		}
	}
	```

modules/network.go
natefikru: I've talked to @twhay and @AusIV, we are moving foreward with your suggestion of setting these "UNHEALTHY" providers to "Warning" status until they hit a threshold of the definied healthCheckThreshold. 

For example: if a provider returns unhealthy for the first time, they are auto updated to warning. if the healthCheckThreshold value == 2, then if the provider returns another unhealthy status for the 2nd time in a row, they will then be marked as unhealthy. The healthCheckThreshold value is configurable on a network basis.

modules/network.go
natefikru: if the provider is behind the blocklag limit, then they are put in the warning status. If they are too far ahead of the blockJumpLimit, then they are put as Unhealthy.

modules/network.go
natefikru: i removed the monotonic check.

modules/network.go
natefikru: im using the appropriate chainID equivalents for solana and bitcoin. Other blockchains will need to be evaluated and configured for as they.

PR #79, opened 2025-02-12
Feature/robust health checks with monodin

natefikru:
# Enhanced Health Check System and E2E Testing Improvements

## Overview
This PR implements a comprehensive health check system for the DIN Caddy proxy, along with improvements to the E2E testing infrastructure. The health check system now monitors provider health more intelligently, with better detection of stalled providers, chain ID verification, and archive mode support.

### Health Check System
- Implemented a robust block history tracking system for each provider
- Added chain ID verification to ensure providers are serving the correct blockchain network
- Added support for archive mode detection for historical queries
- Improved stalled provider detection with configurable thresholds
- Added intelligent handling of network-wide outages vs. individual provider issues
- Updated health status levels (Healthy, Warning, Unhealthy) with clear criteria

### E2E Testing Improvements
- Added a health check endpoint to the test Caddyfile
- Enhanced the request script with better error handling and server readiness checks
- Added proper waiting for server startup before sending test requests
- Improved response logging for better debugging

### Configuration Enhancements
- Added new configuration options in the Caddyfile:
  - `chain_id`: Required parameter for blockchain network identification
  -  `chainid_method`: Required if for non evm networks
  - `block_jump_limit`: Detects providers reporting blocks too far ahead
  - `block_history_size`: Controls the number of blocks to keep in history
  - `archive_enabled`: Enables archive mode checking
  - `call_contract_method` : Required if for non evm networks that have archive enabled

### Documentation
- Added comprehensive documentation in modules/README.md explaining the health check system
- Updated the main README.md with health check information

### Other Improvements
- Added HTTP client timeout to prevent hanging requests
- Simplified Prometheus metrics for health checks
- Updated dependencies, including stretchr/testify to v1.10.0
- Expanded test coverage for the new health check functionality

## Testing
The changes have been tested with the enhanced E2E test suite, which now includes better error handling and more comprehensive checks. All tests are passing.

### DIFF
--- a/modules/din_middleware_helpers.go
+++ b/modules/din_middleware_helpers.go
@@ -205,53 +205,111 @@ func (d *DinMiddleware) updateNetworkWithRegistryData(regNetwork *din.Network, n
 
 // syncNetworkConfig updates the network object with the registry network config data
 func (d *DinMiddleware) syncNetworkConfig(regNetwork *din.Network, network *network) (*network, error) {
+	// Get the healthcheck method name from the registry, usually eth_blockNumber or similar
 	registryHCMethod, err := d.DingoClient.GetNetworkMethodNameByBit(regNetwork.Name, regNetwork.NetworkConfig.HealthcheckMethodBit)
 	if err != nil {
 		d.logger.Error("Failed to get network healthcheck method name", zap.String("network", regNetwork.Name), zap.Error(err))
 		return nil, err
 	}
 
-	registryChainIDMethod, err := d.DingoClient.GetChainIDMethodByBit(regNetwork.Name, regNetwork.NetworkConfig.ChainIDMethodBit)
+	// Get the chain ID method name from the registry, usually eth_chainId or similar
+	registryChainIdMethod, err := d.DingoClient.GetNetworkMethodNameByBit(regNetwork.Name, regNetwork.NetworkConfig.ChainIdMethodBit)
 	if err != nil {
 		d.logger.Error("Failed to get network chain ID method name", zap.String("network", regNetwork.Name), zap.Error(err))
 		return nil, err
 	}
 
-	// Sync chain ID if provided in registry
-	registryChainID := regNetwork.NetworkConfig.ChainID
-	if registryChainID != "" && registryChainID != network.ChainID {
-		d.logger.Debug("Setting network chain ID", zap.String("network", network.Name), zap.String("chain_id", registryChainID))
-		network.ChainID = registryChainID
+	// Get the call contract method name from the registry, usually eth_call or similar
+	registryCallContractMethod, err := d.DingoClient.GetNetworkMethodNameByBit(regNetwork.Name, regNetwork.NetworkConfig.CallContractMethodBit)
+	if err != nil {
+		d.logger.Error("Failed to get network call contract method name", zap.String("network", regNetwork.Name), zap.Error(err))
+		return nil, err
 	}
-	// Sync the value if it is not 0 and different from the current middleware network value
+
+	// Update Chain ID if changed
+	if regNetwork.NetworkConfig.ChainId != "" && regNetwork.NetworkConfig.ChainId != network.ChainId {
+		d.logger.Debug("Setting network chain Id",
+			zap.String("network", network.Name),
+			zap.String("chain_id", regNetwork.NetworkConfig.ChainId))
+		network.ChainId = regNetwork.NetworkConfig.ChainId
+	}
+
+	// Update Healthcheck Method if changed
 	if registryHCMethod != "" && registryHCMethod != network.HCMethod {
-		d.logger.Debug("Setting network healthcheck method", zap.String("network", network.Name), zap.String("method", registryHCMethod))
+		d.logger.Debug("Setting network healthcheck method",
+			zap.String("network", network.Name),
+			zap.String("healthcheck_method", registryHCMethod))
 		network.HCMethod = registryHCMethod
 	}
-	registryHCInterval := int(regNetwork.NetworkConfig.HealthcheckIntervalSec)
-	if registryHCInterval != 0 && registryHCInterval != network.HCInterval {
-		d.logger.Debug("Setting network healthcheck interval", zap.String("network", network.Name), zap.Int("interval", registryHCInterval))
-		network.HCInterval = registryHCInterval
+
+	// Update Chain ID Method if changed
+	if registryChainIdMethod != "" && registryChainIdMethod != network.ChainIdMethod {
+		d.logger.Debug("Setting network chain ID method",
+			zap.String("network", network.Name),
+			zap.String("chain_id_method", registryChainIdMethod))
+		network.ChainIdMethod = registryChainIdMethod
 	}
-	registryBlockLagLimit := int64(regNetwork.NetworkConfig.BlockLagLimit)
-	if registryBlockLagLimit != 0 && registryBlockLagLimit != network.BlockLagLimit {
-		d.logger.Debug("Setting network block lag limit", zap.String("network", network.Name), zap.Int64("block_lag_limit", registryBlockLagLimit))
-		network.BlockLagLimit = int64(registryBlockLagLimit)
+
+	// Update Call Contract Method if changed
+	if registryCallContractMethod != "" && registryCallContractMethod != network.CallContractMethod {
+		d.logger.Debug("Setting network call contract method",
+			zap.String("network", network.Name),
+			zap.String("call_contract_method", registryCallContractMethod))
+		network.CallContractMethod = registryCallContractMethod
 	}
-	registryBlockJumpLimit := int64(regNetwork.NetworkConfig.BlockJumpLimit)
-	if registryBlockJumpLimit != 0 && registryBlockJumpLimit != network.BlockJumpLimit {
-		d.logger.Debug("Setting network block jump limit", zap.String("network", network.Name), zap.Int64("block_jump_limit", registryBlockJumpLimit))
-		network.BlockJumpLimit = int64(registryBlockJumpLimit)
+
+	// Update Healthcheck Interval if changed
+	hcInterval := int(regNetwork.NetworkConfig.HealthcheckIntervalSec)
+	if hcInterval != 0 && hcInterval != network.HCInterval {
+		d.logger.Debug("Setting network healthcheck interval",
+			zap.String("network", network.Name),
+			zap.Int("interval", hcInterval))
+		network.HCInterval = hcInterval
 	}
-	registryMaxRequestPayloadSizeKB := int64(regNetwork.NetworkConfig.MaxRequestPayloadSizeKb)
-	if registryMaxRequestPayloadSizeKB != 0 && registryMaxRequestPayloadSizeKB != network.MaxRequestPayloadSizeKB {
-		d.logger.Debug("Setting network max request payload size", zap.String("network", network.Name), zap.Int64("max_request_payload_size_kb", registryMaxRequestPayloadSizeKB))
-		network.MaxRequestPayloadSizeKB = registryMaxRequestPayloadSizeKB
+
+	// Update Block Lag Limit if changed
+	blockLagLimit := int64(regNetwork.NetworkConfig.BlockLagLimit)
+	if blockLagLimit != 0 && blockLagLimit != network.BlockLagLimit {
+		d.logger.Debug("Setting network block lag limit",
+			zap.String("network", network.Name),
+			zap.Int64("block_lag_limit", blockLagLimit))
+		network.BlockLagLimit = blockLagLimit
 	}
-	registryRequestAttemptCount := int(regNetwork.NetworkConfig.RequestAttemptCount)
-	if registryRequestAttemptCount != 0 && registryRequestAttemptCount != network.RequestAttemptCount {
-		d.logger.Debug("Setting network request attempt count", zap.String("network", network.Name), zap.Int("request_attempt_count", registryRequestAttemptCount))
-		network.RequestAttemptCount = int(registryRequestAttemptCount)
+
+	// Update Block Jump Limit if changed
+	blockJumpLimit := int64(regNetwork.NetworkConfig.BlockJumpLimit)
+	if blockJumpLimit != 0 && blockJumpLimit != network.BlockJumpLimit {
+		d.logger.Debug("Setting network block jump limit",
+			zap.String("network", network.Name),
+			zap.Int64("block_jump_limit", blockJumpLimit))
+		network.BlockJumpLimit = blockJumpLimit
+	}
+
+	// Update Max Request Payload Size if changed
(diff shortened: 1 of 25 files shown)


modules/din_middleware_helpers.go
```diff
@@ -205,36 +205,111 @@ func (d *DinMiddleware) updateNetworkWithRegistryData(regNetwork *din.Network, n
...
+
+	// Get the call contract method name from the registry, usually eth_call or similar
+	registryCallContractMethod, err := d.DingoClient.GetNetworkMethodNameByBit(regNetwork.Name, regNetwork.NetworkConfig.CallContractMethodBit)
+	if err != nil {
+		d.logger.Error("Failed to get network call contract method name", zap.String("network", regNetwork.Name), zap.Error(err))
+		return nil, err
+	}
+
+	// Update Chain ID if changed
+	if regNetwork.NetworkConfig.ChainId != "" && regNetwork.NetworkConfig.ChainId != network.ChainId {
```
ggarri, line 230: According to the settings, we dont allow empty chainID ([see here](https://github.com/DIN-center/din-caddy-plugins/pull/79/files#diff-efc85621f78114184ee60a1ce992a1acd3c7e363ab753b143b16c2258f575ceeR503)), why adding the control over if at this IF statement?

modules/network.go
```diff
@@ -81,218 +84,349 @@ func (n *network) startHealthcheck() {
...
-	return values, ok
-}
+func (n *network) getLatestBlockNumber(httpUrl string, headers map[string]string, ac auth.IAuthClient) (int64, HealthStatus, error) {
+	var lastErr error
+	var lastHealthStatus HealthStatus = Unhealthy
+
+	// Layer 1: Handle attempts
+	for attempt := 0; attempt < n.RequestAttemptCount; attempt++ {
+		resBytes, statusCode, err := n.tryBlockNumberRequestWithTimeout(httpUrl, headers, ac)
+		if err != nil {
```
ggarri, line 344:
```suggestion
		if err != nil && lastHealthStatus == Unhealthy{
```

IMO if we got an error in processBlockNumberResponse which leaded to an updated status, better to hold that message, instead of overwritting it over next attempt, so returned status and error message belong to the same iteration of the loop

modules/network.go
```diff
@@ -81,218 +84,349 @@ func (n *network) startHealthcheck() {
...
-	checkedProviders := n.CheckedProviders
-	for providerName, healthCheckList := range checkedProviders {
-		if healthCheckList[0].blockNumber+n.BlockLagLimit < n.latestBlockNumber {
-			n.Providers[providerName].markWarning()
+// Layer 2: Handle timeout
+func (n *network) tryBlockNumberRequestWithTimeout(httpUrl string, headers map[string]string, ac auth.IAuthClient) ([]byte, *int, error) {
+	payload := []byte(fmt.Sprintf(`{"jsonrpc":"2.0","method": "%s","params":[],"id":1}`, n.HCMethod))
+
+	// Try the request up to 2 times if it times out
+	for retryCount := 0; retryCount < 2; retryCount++ {
```
ggarri: As I see, we already do the retries on the caller of this function `getLatestBlockNumber()`, so I see no reason for this additional complexity

modules/network.go
```diff
@@ -81,218 +84,349 @@ func (n *network) startHealthcheck() {
...
+		type result struct {
+			resBytes   []byte
+			statusCode *int
+			err        error
+		}
+		resChan := make(chan result, 1)
+
+		// Make the request in a goroutine
+		go func(payload []byte) {
+			resBytes, statusCode, err := n.HttpClient.Post(httpUrl, headers, payload, ac)
```
ggarri: To avoid this complex block I would rather recommend to modify the `(h *HTTPClient) Post(..)`
to accept context as argument and then create a `http.RequestWithContext()` so that, if you decide to still retry, you might only retry on context cancel instead of doing it on every error types

modules/network.go
```diff
@@ -301,65 +435,82 @@ func (n *network) getLatestBlockNumber(httpUrl string, headers map[string]string
...
+
+	return fullChainId, nil
 }
 
-// getPercentileBlockNumber returns the block number at the specified percentile across all providers
-// percentile should be between 0 and 1 (e.g., 0.75 for 75th percentile)
-func (n *network) getPercentileBlockNumber(percentile float64) int64 {
-	if len(n.Providers) == 0 {
-		return 0
+func (n *network) testArchiveMode(httpUrl string, headers map[string]string, ac auth.IAuthClient, quarterBlockHeightHex string) error {
```
ggarri: what is this test evaluating?

modules/network.go
```diff
@@ -81,218 +84,349 @@ func (n *network) startHealthcheck() {
...
+	case string:
+		if result == "" || result[:2] != "0x" {
+			return 0, Unhealthy, errors.New("Invalid block number")
+		}
+
+		blockNumber, err = strconv.ParseInt(result[2:], 16, 64)
+		if err != nil {
+			return 0, Unhealthy, errors.Wrap(err, "Error converting block number")
+		}
+	case float64:
```
ggarri, line 391: is there not cases where blockNumber comes back as a number between quotes? I am trying to understand in which cases we get `float64`

modules/network.go
natefikru: renamed

modules/network.go
natefikru: default decoded value for json numbers on solana and starknet return back as float64 when the struct is labeled as an interface.

Im not aware of any networks that return back block numbers with quotes.

modules/network.go
natefikru: i took this out the redundant retry. and added the timeout in the http client level

modules/network.go
natefikru: removed this logic fully, updated the logic to the following:

```
func NewHTTPClient() *HTTPClient {
	return &HTTPClient{
		httpClient: &http.Client{
			Timeout: 5 * time.Second, // Set timeout to 5 seconds
			Transport: &http.Transport{
				Proxy: http.ProxyFromEnvironment,
				DialContext: (&net.Dialer{
					Timeout:   30 * time.Second,
					KeepAlive: 30 * time.Second,
				}).DialContext,
				MaxIdleConnsPerHost:   16,
				MaxIdleConns:          16,
				IdleConnTimeout:       90 * time.Second,
				TLSHandshakeTimeout:   10 * time.Second,
				ExpectContinueTimeout: 1 * time.Second,
			},
		},
	}
}
```

modules/network.go
natefikru: updated name

modules/network.go
natefikru: i removed this function completely.

modules/network.go
```diff
@@ -81,285 +84,443 @@ func (n *network) startHealthcheck() {
...
+				zap.Int64("network_block", latestNetworkBlock))
+			if Warning > worstStatus {
+				worstStatus = Warning
+			}
+		} else if allStalled {
+			// All providers are stalled - likely a network outage, don't change status
+			// This is set to healthy because we want to return traffic as it is in the case of a network outage
+			n.logProviderWarning("All providers are stalled", provider)
+		} else {
+			// This provider is stalled while others are progressing - provider outage
```
AusIV: Is this condition practically reachable? The implication here is that:

1. We are stalled.
2. We are *not* lagged.
3. Other providers are not stalled.

Can we be stalled but not lagged while other all providers are not stalled? Wouldn't us being stalled while other providers are not stalled generally necessitate being lagged? And even if it is reachable, if we're stalled while other providers aren't, but we're still close enough to current to not be considered lagged, why is that unhealthy while stalled *and* lagged is just warning?

I'd be inclined to simplify this whole set of conditions to:

```
if isLagged {
  // Mark warning
  if isStalled {
    // Mark unhealthy
  }
}
```

There's an implicit assumption here that if we're stalled but not lagged, then nobody else is doing significantly better than us. We don't have to explicitly check the other providers, because the fact that we're *not* lagged means the other providers can't have gotten very far ahead of us, and thus are also going to be stalled.

modules/provider.go
```diff
@@ -11,36 +15,53 @@ import (
...
+	logger     *zap.Logger
+	Priority   int
 
 	// Registry Configuration Values
 	Methods []*string            `json:"methods"`
 	Auth    *siwe.SIWEClientAuth `json:"auth"`
 
-	consecutiveHealthyChecks int
+	consecutiveUnhealthyChecks int
+	blockHistory               []blockHistoryEntry
```
noleto: probably a better data structure for `blockHistory` is to use the Go standard double linked list: https://pkg.go.dev/container/list

modules/provider_test.go
```diff
@@ -2,338 +2,424 @@ package modules
...
+			name:            "hex chain ID",
+			chainId:         "0x1",
+			expectedChainId: "0x1",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := tt.chainId
+			assert.Equal(t, tt.expectedChainId, result)
```
noleto: which logic is being tested here? I

noleto (commented):
Thanks for the README doc around the health check; it helped a lot to understand the logic. I left a few comments.

modules/network.go
natefikru: simplified the logic making "stalled and lagged" an unhealthy state is more intuitive than just a warning, and the edge case handling is much cleaner now.

modules/provider_test.go
natefikru: nothing, removed this function, good catch

modules/provider.go
natefikru: updated to list.List and updated tests.

modules/network.go
natefikru:
```
	// Check for stalling
	isStalled := n.isStalled(provider)
	allStalled := n.allProvidersStalled()

	if isLagged {
		// Provider is lagging behind
		n.logProviderWarning("Provider is lagging behind network", provider,
			zap.Int64("block_lag", blockLag),
			zap.Int64("provider_block", currentBlock),
			zap.Int64("network_block", latestNetworkBlock))
		if Warning > worstStatus {
			worstStatus = Warning
		}

		if isStalled {
			// Provider is both stalled and lagged - more serious issue
			n.logProviderWarning("Provider is stalled and lagged", provider)
			return Unhealthy
		}
	} else if isStalled && !allStalled {
		// Edge case: Provider is stalled but not yet lagged, while others are making progress
		n.logProviderWarning("Provider is stalled while others are progressing", provider)
		return Unhealthy
	}
	```

File: modules/network.go [hunk 10]
```diff
@@ -81,285 +84,443 @@ func (n *network) startHealthcheck() {
...
+	return blockNumber, Healthy, nil
 }
 
-// getPercentileBlockNumber returns the block number at the specified percentile across all providers
-// percentile should be between 0 and 1 (e.g., 0.75 for 75th percentile)
-func (n *network) getPercentileBlockNumber(percentile float64) int64 {
-	if len(n.Providers) == 0 {
-		return 0
-	}
+func (n *network) getChainID(httpUrl string, headers map[string]string, ac auth.IAuthClient) (string, error) {
```
cds-amal, line 400: Memoizing this function could save network calls.

cds-amal (commented):
Nice improvements to the health check system! I noticed a memoization opportunity and have a question about the retry implementations in network.go - I notice there are three similar retry loops:

```go
// In getLatestBlockNumber, getChainID, and archiveModeCheck
for attempt := 0; attempt < n.RequestAttemptCount; attempt++ {
    // Make request
    if err != nil {
        lastErr = err
        continue  // Immediate retry
    }
}
```

What happens during network congestion when multiple providers start timing out simultaneously? With health checks running every 5s and each provider getting immediate retries, have you considered the potential impact on system resources?

ggarri (approved):

noleto (approved):
Good to go!

modules/network.go
AusIV: The network calls on getChainID are part of the point. We've had problems in the past with providers' routers starting to send traffic to nodes for other chains, and the easiest / cheapest call we have to uniquely identify what network is being served is getting the chainid.

There might be a case for doing it less often than the block number check, but memoizing would not tell us when a provider starts serving the wrong chain.

modules/network.go
AusIV: Again, I don't think the isStalled && !allStalled check is particularly useful if we're not yet lagged.

If it's a situation where the whole network was stalled for a while and has just started moving again, some other providers are a block or two ahead of us but we haven't processed them yet, but aren't far enough behind to be considered lagged, I don't think that's something we need to mark as unhealthy. To my mind, stalled only merits being unhealthy if we're also lagged.

I'll also note that part of my concern here is that the `n.allProvidersStalled()` check is potentially kind of expensive. It acquires certain locks and does a lot of memory copies to calculate, and the way things are currently structured it would repeat those lock acquisitions and memory copies for each provider every healthcheck. Doing those checks for every provider seems potentially expensive to cut off such a narrow case.

That said, if instead of setting allStalled before the if conditions we did:

```
 else if isStalled && !n.allProvidersStalled() {
```

Then we only execute the allProviderStalled() check if we're not lagged and we are stalled. I still don't necessarily agree with that behavior, but again it's a very narrow case to consider so reducing the computational expense would make me not worried about it.

natefikru:
> Nice improvements to the health check system! I noticed a memoization opportunity and have a question about the retry implementations in network.go - I notice there are three similar retry loops:
> 
> ```go
> // In getLatestBlockNumber, getChainID, and archiveModeCheck
> for attempt := 0; attempt < n.RequestAttemptCount; attempt++ {
>     // Make request
>     if err != nil {
>         lastErr = err
>         continue  // Immediate retry
>     }
> }
> ```
> 
> What happens during network congestion when multiple providers start timing out simultaneously? With health checks running every 5s and each provider getting immediate retries, have you considered the potential impact on system resources?

The provider requests are being done synchronously so im not sure it would have a direct impact if the requests aren't happening at the same time.

modules/network.go
natefikru: ahhh gotcha, i refactored the code to change where the isStalled and allProvidersStalled calls are being done and updated the status.

If you still think it isn't neccesary to include lmk, I'll take it out, otherwise it looks like this:

```
	if isLagged {
		// Provider is lagging behind
		n.logProviderWarning("Provider is lagging behind network", provider,
			zap.Int64("block_lag", blockLag),
			zap.Int64("provider_block", currentBlock),
			zap.Int64("network_block", latestNetworkBlock))
		if Warning > worstStatus {
			worstStatus = Warning
		}

		isStalled := n.isStalled(provider)
		if isStalled {
			// Provider is both stalled and lagged - more serious issue
			n.logProviderWarning("Provider is stalled and lagged", provider)
			return Unhealthy
		}
	} else if n.isStalled(provider) && !n.allProvidersStalled() {
		// Edge case: Provider is stalled but not yet lagged, while others are making progress
		n.logProviderWarning("Provider is stalled while others are progressing", provider)
		return Warning
	}
```

PR #83, opened 2025-03-14
Sync up caddyfile with prod

AusIV:


Caddyfile
```diff
@@ -164,15 +210,26 @@
 							priority 0
 						}
 						https://infura-zksync.liquify.com/api=key {
-							priority 0
+							priority 1
 						}
 					}
+					chain_id eip155:0x144
+					archive_enabled true
```
natefikru: archive should be disabled for zksync-mainnet, unless you just want to double check that it won't work.

Caddyfile
```diff
@@ -213,6 +279,10 @@
 						}
 					}
 					healthcheck_method starknet_blockNumber
+					chainid_method starknet_chainId
+					chain_id starknet:0x534e5f5345504f4c4941
+					call_contract_method starknet_call
+					archive_enabled true
```
natefikru: archive checks for starknet Mainnet + Testnet should be disabled.

Caddyfile
```diff
@@ -255,49 +333,72 @@
...
+							priority 0
+							headers {
+								x-api-key key
+							}
+						}
+					}
+					chain_id eip155:0x82
+					archive_enabled true
 				}
 				bitcoin-mainnet {
```
natefikru, line 401:
```
bitcoin-mainnet {
	providers {
		https://din-router.extrnode.com
	}
	healthcheck_method getblockcount
	chainid_method getblockchaininfo
	chain_id bip:main
}
```

natefikru:
I believe we are not using Caddyfile methods anywhere in the router code, we could probably remove all references to the method list in this file to simplify it.

AusIV:
> I believe we are not using Caddyfile methods anywhere in the router code, we could probably remove all references to the method list in this file to simplify it.

Probably, though I'm not inclined to do that this PR since I'm trying to get it to be a reflection of what's out to beta.

Originally I intended that to be a place to keep track of what methods providers were supposed to be supporting so we'd have it when we needed to start using it, but in practice it hasn't been maintained as we expanded to new networks, so when we need it we'll have to re-evaluate anyway.

PR #89, opened 2025-04-04
NFT support

AusIV:
This implements the provider-side of an NFT authentication protocol.

If a signer is not in the whitelist, it checks the NFT specified in the resources section of the SIWE claim, and if the NFT corresponds to this provider and is owned by the signer, the session token will be issued.

This is a draft, with a few things I know I want to do before it gets merged, but some of those things will involve collaboration, so this seems like a good place for discussion.

- [ ] The NFT Contract is in the [monorepo](https://github.com/DIN-center/monodin/pull/103). We should discuss moving the libraries that interact with it over to monodin as well - I started with them here because this is where I needed them, but I suspect other tools will need to interact with the contract.
- [ ] The NFT Contract needs tests. It's a pretty simple extension of OpenZeppelin's Enumerable ERC721 contract, but has a few things that could be tested.
- [ ] Right now this requires setting a provider_id field in the Caddyfile. I want to work with @natefikru on calculating it in the registry integration.
- [ ] The siwe-signer tool needs to be extended to support NFTs. This might be separable into its own work, because it still works with whitelisted addresses absent NFT support.
- [ ] The auth/siwe package could use some tests that cover NFT support.


lib/auth/siwe/nft_manager.go
```diff
@@ -0,0 +1,109 @@
...
+func (m *NFTManager) Refresh() error {
+
+	tokenIds, err := m.OptionsManager.TokenIDsByOwner(m.Owner)
+	if err != nil {
+		m.logger.Warn("error getting token ids by owner", zap.String("error", err.Error()))
+		return err
+	}
+
+	now := uint64(time.Now().Unix())
+	bufferTime := now + 60
```
natefikru, line 61: 60 should probably be configrable

natefikru:
overall great implementation, all updates look solid. a couple small notes, it looks like redis is added in the go.mod but it doesn't look like it is being used, noticed that ethgo is being downgraded, is there a reason for that? Lastly, it might make sense to add retry logic in the ethgo request process flow.  

we should probably have some unit additional tests in the auth/siwe folder for the additional logic.

PR #91, opened 2025-04-10
Enable naive method-based routing

AusIV:
This implements a very simple version of method based routing. Networks can specify a list of `routed_methods`, and providers can specify a list of `methods`.

If the network specifies a list of `routed_methods`, any request that has a method in the `routed_methods` list will only be eligible to be routed to providers that specify the same method in `methods`.

This does very little to ensure consistency. It will make sure that method routed requests with same Din-Session-Id will go to the same provider for the same method, but that may be a different provider than non-method-routed requests, so they may be on different blocks.

### DIFF
--- a/modules/din_provider_filter.go
+++ b/modules/din_provider_filter.go
@@ -7,10 +7,19 @@ import (
 	"github.com/caddyserver/caddy/v2"
 )
 
+
+// ProviderFilter offers an interface for determining which providers can handle
+// a particular request. This is similar to what happens in the upstreams module,
+// but lets us attach filtering logic to specific networks.
 type ProviderFilter interface {
     FilterProviders(*http.Request, map[string]*provider) map[string]*provider
 }
 
+
+// methodFilter implements the ProviderFilter. If a network needs to route specific
+// methods to a subset of providers, they can indicate the method's importance here,
+// and the methodFilter will compare against the providers' method sets to decide where
+// to route.
 type methodFilter struct {
 	FilteredMethods map[string]struct{}
 }
@@ -43,6 +52,7 @@ func (mf *methodFilter) FilterProviders(r *http.Request, p map[string]*provider)
 	}
 	for k, provider := range p {
 		if _, ok := provider.Methods[request.Method]; ok {
+			// This provider supports the specified filtered method, and is elligible to serve the request
 			result[k] = provider
 		}
 	}
--- a/modules/din_provider_filter.go
+++ b/modules/din_provider_filter.go
@@ -1,0 +1,54 @@
+package modules
+
+import (
+	"encoding/json"
+	"net/http"
+	din_http "github.com/DIN-center/din-caddy-plugins/lib/http"
+	"github.com/caddyserver/caddy/v2"
+)
+
+type ProviderFilter interface {
+    FilterProviders(*http.Request, map[string]*provider) map[string]*provider
+}
+
+type methodFilter struct {
+	FilteredMethods map[string]struct{}
+}
+
+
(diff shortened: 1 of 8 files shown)


modules/din_provider_filter.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		// Body isn't JSON, send it to any provider and good luck
+		return p
+	}
+	if _, ok := mf.FilteredMethods[request.Method]; !ok {
+		// Method isn't an important one, send to any provider
+		return p
+	}
+	for k, provider := range p {
+		if _, ok := provider.Methods[request.Method]; ok {
```
natefikru, line 54: a comment here would help make this more clear

modules/din_provider_filter.go
```diff
@@ -0,0 +1,54 @@
...
+			result[k] = provider
+		}
+	}
+	if len(result) == 0 {
+		// Nothing supports this method. Send it to any provider and good luck
```
natefikru, line 60: might be a good idea to have a debug log here.

modules/network.go
```diff
@@ -29,6 +29,8 @@ type network struct {
 	HCThreshold      int
 	BlockHistorySize int
 
+	MethodFilter     *methodFilter
```
natefikru, line 34: add a comment explaining

modules/din_provider_filter.go
```diff
@@ -0,0 +1,54 @@
...
+	"net/http"
+	din_http "github.com/DIN-center/din-caddy-plugins/lib/http"
+	"github.com/caddyserver/caddy/v2"
+)
+
+type ProviderFilter interface {
+    FilterProviders(*http.Request, map[string]*provider) map[string]*provider
+}
+
+type methodFilter struct {
```
natefikru, line 23: comments explaining how the provider filter works would be very helpful.

as well as an addition in the docs

natefikru:
I think in general additional comments and documentation would be very much needed here. 

regarding the lack of consistency, i also think more of a description should be found as well.

additional validation would help as well to make sure that requests are not malformed, potentially could be added in a v2 update

modules/din_provider_filter.go
AusIV: I'm inclined to put a warning log in the provision step. I don't think we need to log every time a request comes through, but having a log during provisioning if it's not configured properly seems prudent.

AusIV:
@natefikru 

I've updated to address most of your concerns. I'm not sure what you mean by:

> additional validation would help as well to make sure that requests are not malformed, potentially could be added in a v2 update

We route the requests we get from clients. If we can't parse the request we can't route based on details of the request, and we can generally expect that the request will fail when it gets to a provider, but I think it's generally best to let a provider try to process the request and reply with an error message consistent with that network's clients. It could get very cumbersome for us if we try to avoid routing malformed requests *and* respond as each network's clients would respond.

natefikru:
> @natefikru
> 
> I've updated to address most of your concerns. I'm not sure what you mean by:
> 
> > additional validation would help as well to make sure that requests are not malformed, potentially could be added in a v2 update
> 
> We route the requests we get from clients. If we can't parse the request we can't route based on details of the request, and we can generally expect that the request will fail when it gets to a provider, but I think it's generally best to let a provider try to process the request and reply with an error message consistent with that network's clients. It could get very cumbersome for us if we try to avoid routing malformed requests _and_ respond as each network's clients would respond.

good point on that. maybe we can add the validation on the provider's input form on the client.

natefikru (approved):

PR #94, opened 2025-04-15
Removed the getting of the latest healthy block in archive check

natefikru:


AusIV (approved):

AusIV:
Interesting. This is where Github's diffs can hide important things. If I'd realized that there was already a `currentBlock` in scope I'd have asked why we were making a new one. Just seeing the diff that didn't include the other one that never occurred to me.

Approved.

PR #104, opened 2025-05-06
Feature: Network Loopback Health Checks & Metrics

natefikru:
This PR introduces self-health checks for the router's own network endpoints and adds dedicated Prometheus metrics for monitoring their performance and status.

**Key Changes:**

*   **Network Self-Health Check:**
    *   Added `checkSelfLoopbackHealth()` in `modules/network.go` to perform asynchronous loopback requests to `http://localhost:{port}/{network.Name}`.
    *   Fixed a nil pointer dereference related to `statusCode` handling.
*   **New Prometheus Metrics (`lib/prometheus/`):**
    *   `DinNetworkHealthCheckCount`: Counts network-level health checks.
    *   `DinNetworkRequestHealthCheckDurationMilliseconds`: Tracks duration of these checks.
    *   Updated interfaces, mocks, and added new tests.
*   **Metric Renaming:**
    *   Provider-specific health metrics renamed from `DinHealthCheck...` to `DinProviderHealthCheck...` for clarity.

**Benefits:**
Improves observability by distinguishing router-level issues from provider issues and provides dedicated metrics for router endpoint health.


lib/prometheus/prometheus.go
```diff
@@ -69,23 +73,43 @@ func RegisterMetrics() {
...
-	prometheus.MustRegister(DinRequestCount, DinHealthCheckCount, DinRequestDurationMilliseconds, DinRequestBodyBytes, DinHealthCheckBlockNumber)
+	prometheus.MustRegister(DinRequestCount, DinProviderHealthCheckCount, DinRequestDurationMilliseconds, DinRequestBodyBytes, DinProviderHealthCheckBlockNumber)
+
+	// Register network level health check metrics
+	DinNetworkHealthCheckCount = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "din_network_health_check_count",
+			Help: "Metric for counting network-level health checks",
+		},
+		[]string{"network", "response_status", "machine_id", "environment"},
```
noleto: Should we also send the `"service"` tag to enable filtering in the dashboard?

lib/prometheus/prometheus.go
```diff
@@ -69,23 +73,43 @@ func RegisterMetrics() {
...
+	)
+
+	DinNetworkRequestHealthCheckDurationMilliseconds = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "din_network_request_health_check_duration_milliseconds",
+			Help:    "Metric for measuring the duration of network-level health checks",
+			Buckets: []float64{1, 5, 10, 25, 50, 100, 250, 500, 1000, 2500, 5000, 10000, 30000, 60000}, // 1ms to 60s (1 minute)
+		},
+		[]string{"network", "response_status", "machine_id", "environment"},
```
noleto: same, also send the `"service"` tag to enable filtering in the dashboard?

modules/network.go
```diff
@@ -93,6 +93,43 @@ func (n *network) startHealthcheck() {
 
 // HealthCheck performs health checks on all providers and updates their status
 func (n *network) healthCheck() {
+	// Self loopback health check (run asynchronously)
```
noleto, line 98: This function is starting to get very lengthy and now handles two responsibilities: checking all providers and performing self-checks at the network level. I think it would improve readability if it were split into two separate functions, like this:

```go
func (n *network) healthCheck() {
    go n.LoopbackHealthCheck()

    n.ProvidersHealthCheck()
}
```

modules/network.go
```diff
@@ -645,3 +682,45 @@ func (n *network) hasOtherHealthyProviders(provider *provider) bool {
 	}
 	return false
 }
+
+// checkSelfLoopbackHealth performs a health check on the router's own endpoint (loopback)
+func (n *network) checkSelfLoopbackHealth() (*getLatestBlockNumberResult, error) {
+	url := fmt.Sprintf("http://localhost:8000/%s", n.Name)
```
noleto: Hardcoding `http://localhost:8000/` seems a bit inflexible, right? What if the service is deployed on a different port? Perhaps we could read the configuration from the Caddyfile, as we're checking against the same service/process.

modules/network.go
```diff
@@ -645,3 +682,45 @@ func (n *network) hasOtherHealthyProviders(provider *provider) bool {
...
+	// Use the payload for getLatestBlockNumber
+	payload := []byte(fmt.Sprintf(`{"jsonrpc":"2.0","method": "%s","params":[],"id":1}`, n.HCMethod))
```
noleto: Since `"params":[]` are optional in an EVM-compatible blockchain, it’s probably better to remove them to avoid errors when interacting with non-EVM blockchains like Solana

Additionally, this payload should be overridden per network, as it wouldn’t work for blockchains like Bitcoin or Starknet for instance.

noleto (commented):
Looks good to me overall. I left a few comments to discuss, but probably nothing that would prevent us from giving it a try.  

Also, I'm wondering what is the main goal of this "self-check"? Is it to trigger an alerting tool based on the metrics sent to Prometheus?

modules/network.go
natefikru: updated. it technically does work for starknet btw :)

modules/network.go
natefikru: updated

modules/network.go
natefikru: updated

lib/prometheus/prometheus.go
natefikru: updated

lib/prometheus/prometheus.go
natefikru: updated

natefikru:
> Looks good to me overall. I left a few comments to discuss, but probably nothing that would prevent us from giving it a try.
> 
> Also, I'm wondering what is the main goal of this "self-check"? Is it to trigger an alerting tool based on the metrics sent to Prometheus?

Thanks for the review leo, yes exactly, it is to trigger alerting based on the overall health of an endpoint, similar to what infura see's today.

noleto (approved):
LGTM!

modules/network.go
```diff
@@ -659,3 +701,48 @@ func (n *network) hasOtherHealthyProviders(provider *provider) bool {
 	}
 	return false
 }
+
+// checkSelfLoopbackHealth performs a health check on the router's own endpoint (loopback)
+func (n *network) checkSelfLoopbackHealth() (*getLatestBlockNumberResult, error) {
+	if n.CaddyPort == "" {
+		return nil, errors.New("Caddy port is not set")
+	}
+	url := fmt.Sprintf("http://localhost:%s/%s", n.CaddyPort, n.Name)
```
AusIV: I would point this to `127.0.0.1` instead of `localhost`. On some systems, localhost will get bound to the IPV6 loopback port, which we don't normally configure caddy to listen on.

AusIV (approved):

modules/network.go
natefikru: done.

PR #108, opened 2025-05-14
Feature/provider docs

AusIV:


docs/initial_design.md
```diff
@@ -0,0 +1,34 @@
...
+- A separate agent was required for monitoring providers, as Traefik’s Yaegi interpreter was not capable of maintaining websocket connections to monitor providers.
+- Traefik’s load balancer was positioned after the middleware, leaving no way for the plugin to alter requests after the target provider was known.
+- Traefik’s plugin system offered no way for plugins to communicate with each other, so the provider plugin could not pass information to a middleware plugin, for example.
+- Traefik’s load balancer offered no way to use different paths for different providers. This could be worked around by load balancing across loopback ports on Traefik and manipulating the request with middleware, but that limited the load balancer’s ability to evaluate target health properly.
+
+## Caddy
+
+Caddy's Plugin system is far more detailed than Traefik's, offering plugin hooks in many places throughout the request process. We implement a set of Caddy plugins using the following hooks.
+
+### http.handlers
```
natefikru, line 16: just want to double check that these are the only 3 caddy plugins we are using. should we say anything about our custom filter plugin?

natefikru (approved):

docs/initial_design.md
AusIV: We should, but that part of the docs was actually written before we had the filter plugins and just hadn't been merged yet. I'll make a note on it.

PR #109, opened 2025-05-21
hotfix: add retry error failures

natefikru:
# Refactor: Enhance Request Processing with JSON-RPC Error Detection and Intelligent Retry Logic

This PR introduces comprehensive improvements to request processing, retry logic, and error handling with a focus on JSON-RPC error detection and intelligent retry strategies:

## JSON-RPC Error Detection and Handling
- **JSON-RPC Error Detection**: Added support for detecting JSON-RPC errors in HTTP 200 responses through new `JSONRPCError` and `JSONRPCResponse` types in `lib/http/types.go`
- **Smart Error Classification**: Implemented `checkForJSONRPCError()` function to parse response bodies and identify JSON-RPC errors even when HTTP status is 200
- **Intelligent Retry Logic**: Added `isJSONRPCErrorRetryable()` function that distinguishes between:
  - **Non-retryable errors** (client/request issues): Parse errors, invalid requests, method not found, invalid params
  - **Retryable errors** (server/network issues): Internal errors, server error range (-32000 to -32099), timeouts, connection issues
  - **Message-based detection** for custom error codes using pattern matching
  - **Default to retryable** for unknown error codes to avoid missing transient issues

## Enhanced Retry and Logging System
- **Asynchronous Error Logging**: Failed request attempts are now logged asynchronously via `logFailedAttemptAsync()` to prevent logging operations from adding latency to the main request path
- **Centralized Logging Logic**: Extracted detailed error logging into a dedicated helper function in `din_middleware_helpers.go` for better code organization
- **Unified Logging Approach**: Consolidated duplicate logging calls into a single, comprehensive logging mechanism that handles both HTTP errors and JSON-RPC errors
- **Enhanced Log Context**: Logs now include JSON-RPC error details (code, message, data) when applicable, providing better diagnostic information

## Performance and Reliability Improvements
- **Minimal Performance Impact**: Benchmarking shows JSON-RPC error checking adds only ~674 nanoseconds per request
- **Increased HTTP Timeout**: Extended HTTP client timeout from 5 seconds to 5 minutes to accommodate slower upstream responses
- **Improved Error Handling**: Added proper error logging for request body parsing and response writing failures
- **Updated Retry Comments**: Enhanced documentation explaining when retries occur (HTTP errors vs retryable JSON-RPC errors)

## Testing and Code Quality
- **Comprehensive Unit Tests**: Added extensive test coverage for:
  - `checkForJSONRPCError()` with various response types and edge cases
  - `isJSONRPCErrorRetryable()` covering all error codes and message patterns
  - `logFailedAttemptAsync()` with different parameter types and error scenarios
- **Updated Test Expectations**: Modified existing tests to reflect the change from error-level to warning-level logging for retry attempts
- **Eliminated Code Duplication**: Removed redundant logging and error checking throughout the retry logic

## Breaking Changes
- **Log Level Changes**: Retry attempt logs are now at warning level instead of error level
- **Removed Redundant Logging**: Eliminated duplicate final error logging since all failures are now logged during retry attempts

These changes significantly improve the system's ability to handle JSON-RPC specific errors intelligently, reduce unnecessary retries for client errors, and provide better observability while maintaining optimal performance.

### DIFF
--- a/modules/din_middleware.go
+++ b/modules/din_middleware.go
@@ -338,24 +338,18 @@ func (d *DinMiddleware) ServeHTTP(rw http.ResponseWriter, r *http.Request, next 
 			break
 		}
 
-		// Log an error if this attempt failed and a retry will occur
-		if attempt < networkObj.RequestAttemptCount-1 { // Check if more retries are pending
-			// Call the helper function as a goroutine
-			// Data extraction for logging is now handled within logFailedRetryAttemptAsync
-			go logFailedRetryAttemptAsync(
-				d.logger, // Pass the logger client
-				networkPath,
-				attempt+1, // Attempt number is 1-indexed for logging
-				networkObj.RequestAttemptCount,
-				rww.statusCode,
-				err, // upstream error from next.ServeHTTP
-				repl,
-				requestBody, // Pass the parsed request body (can be nil)
-			)
-		}
-
-		// If the attempt (0-indexed) fails, log the failure and that we are retrying
-		d.logger.Debug("Retrying request", zap.String("network", networkPath), zap.Int("failedAttemptIndex", attempt), zap.Int("status", rww.statusCode))
+		// Call the helper function as a goroutine
+		// Data extraction for logging is now handled within logFailedRetryAttemptAsync
+		go logFailedRetryAttemptAsync(
+			d.logger, // Pass the logger client
+			networkPath,
+			attempt+1, // Attempt number is 1-indexed for logging
+			networkObj.RequestAttemptCount,
+			rww.statusCode,
+			err, // upstream error from next.ServeHTTP
+			repl,
+			requestBody, // Pass the parsed request body (can be nil)
+		)
 	}
 	if err != nil {
 		return errors.Wrap(err, "Error serving HTTP")
--- a/modules/din_middleware.go
+++ b/modules/din_middleware.go
@@ -3,7 +3,6 @@ package modules
 import (
 	"bytes"
 	"encoding/hex"
-	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -315,6 +314,10 @@ func (d *DinMiddleware) ServeHTTP(rw http.ResponseWriter, r *http.Request, next 
 	reqStartTime := time.Now()
 
 	// Retry the request if it fails up to the max attempt request count
+	// Retries occur when:
+	// 1. HTTP errors (non-200 status codes or upstream errors)
+	// 2. Retryable JSON-RPC errors (server errors, timeouts, rate limits, etc.)
+	// Non-retryable JSON-RPC errors (method not found, invalid params) will not trigger retries
 	for attempt := 0; attempt < networkObj.RequestAttemptCount; attempt++ {
 		rww = NewResponseWriterWrapper(rw)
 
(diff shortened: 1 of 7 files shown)


modules/din_middleware.go
```diff
@@ -337,8 +337,25 @@ func (d *DinMiddleware) ServeHTTP(rw http.ResponseWriter, r *http.Request, next
...
+				networkObj.RequestAttemptCount,
+				rww.statusCode,
+				err, // upstream error from next.ServeHTTP
+				repl,
+				requestBody, // Pass the parsed request body (can be nil)
+			)
+		}
+
+		// If the attempt (0-indexed) fails, log the failure and that we are retrying
+		d.logger.Debug("Retrying request", zap.String("network", networkPath), zap.Int("failedAttemptIndex", attempt), zap.Int("status", rww.statusCode))
```
noleto: Shouldn't we be logging the error regardless of whether a retry is coming? At this point, the request has already failed, right?

noleto: `d.logger.Debug` still necessary?

modules/din_middleware_helpers.go
```diff
@@ -1,15 +1,98 @@
...
+	failedAttemptNumber int, // 1-indexed
+	maxAttempts int,
+	statusCodeOfFailure int,
+	upstreamErr error, // Can be nil
+	repl *caddy.Replacer, // Added: Caddy replacer to get context data
+	parsedReqBody *din_http.JSONRPCRequest, // Added: Parsed request body
+) {
+	// --- Extract data within the async function ---
+	provider := "unknown"
+	if provVal, provOk := repl.Get(RequestProviderKey); provOk {
```
noleto, line 31: Is this thread-safe? I noticed that `ServeHTTP(rw http.ResponseWriter, r *http.Request, next caddyhttp.Handler)` was previously protected by a mutex. Now that the log extraction has been moved to an asynchronous function, shouldn't this also be protected to ensure thread safety?

natefikru: can you point to where in code there used to be a mutex via github link?

modules/din_middleware.go
natefikru: removed.

natefikru: updated to always log the error and renamed function as well

modules/din_middleware_helpers.go
noleto: https://github.com/DIN-center/din-caddy-plugins/blob/6b8a7df7b70fb76a4f471b50d22c562641e9d792/modules/din_middleware.go#L251-L253

natefikru: Ah i see now.

The async function uses copies of the critical data or thread-safe objects so we won't have to pass on the mutex itself in the function

modules/din_middleware.go
```diff
@@ -370,9 +381,9 @@ func (d *DinMiddleware) ServeHTTP(rw http.ResponseWriter, r *http.Request, next
 
 			errUnmarshal := json.Unmarshal(bodyDataForErrorLog, &requestForErrorLog)
 			if errUnmarshal != nil {
-				d.logger.Warn("Failed to unmarshal request body for error logging", zap.String("request_body_snippet", string(bodyDataForErrorLog[:min(len(bodyDataForErrorLog), 100)])), zap.String("network", networkPath), zap.String("provider", provider), zap.Int("status", rww.statusCode))
+				d.logger.Error("Failed to unmarshal request body for error logging", zap.String("request_body_snippet", string(bodyDataForErrorLog[:min(len(bodyDataForErrorLog), 100)])), zap.String("network", networkPath), zap.String("provider", provider), zap.Int("status", rww.statusCode))
```
AusIV: I would contend that this should stay warning level. To me error level logs imply that the software has done something wrong, not just that a user provided bad input.

If we're changing this because we need it to be visible, I'd advocate changing the log level for logs to include warnings (though I think we're currently doing INFO level logs, so I don't think this is it)

modules/din_middleware_helpers.go
AusIV: Also the replacer is thread safe: https://github.com/caddyserver/caddy/blob/master/replacer.go

modules/din_middleware_helpers.go
```diff
@@ -1,15 +1,98 @@
...
+			lg.Debug("Async retry error log: Failed to unmarshal requestParams for structured logging, logging as raw string.",
+				zap.Error(errUnmarshal),
+				zap.String("rawParamsAttempted", string(requestParams)))
+		}
+	} else if requestMethod == "" && rawRequestBodySnippet != "" {
+		// Only log rawRequestBodySnippet if we didn't even have a requestMethod
+		logFields = append(logFields, zap.String("rawRequestBodySnippet", rawRequestBodySnippet))
+	}
+
+	lg.Error("Request attempt failed, initiating retry", logFields...)
```
AusIV: Again, I'd consider this a warning level log.

AusIV (approved):

modules/din_middleware.go
natefikru: gotcha, we should update to Warning then for parsed errors.

modules/din_middleware_helpers.go
natefikru: reduced to warning.

AusIV (approved):

PR #124, opened 2025-07-30
Enable OpenID Connect authentication middleware

AusIV:
This is needed to support Esplora APIs in the short term, but might be applicable in other situations later.

### DIFF
--- a/lib/openidconnectproxy/auth.go
+++ b/lib/openidconnectproxy/auth.go
@@ -1,0 +1,139 @@
+package openidconnectproxy
+
+import (
+    // "context"
+    "encoding/json"
+    "net/http"
+    "net/url"
+    "sync"
+    "time"
+
+    "github.com/caddyserver/caddy/v2"
+    "github.com/caddyserver/caddy/v2/caddyconfig/caddyfile"
+    "github.com/caddyserver/caddy/v2/caddyconfig/httpcaddyfile"
+    "github.com/caddyserver/caddy/v2/modules/caddyhttp"
+)
+
+type OpenIDConnectProxy struct {
+    ClientID     string `json:"client_id"`
+    ClientSecret string `json:"client_secret"`
+    TokenURL     string `json:"token_url"`
+    Scope        string `json:"scope"`
+    Upstream     string `json:"upstream"`
+
+    tokenMu      sync.RWMutex
+    accessToken  string
+    tokenExpiry  time.Time
+}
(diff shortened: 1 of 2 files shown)


lib/openidconnectproxy/auth.go
```diff
@@ -0,0 +1,139 @@
...
+    if err != nil {
+        return nil, err
+    }
+    return &m, nil
+}
+
+func init() {
+    caddy.RegisterModule(OpenIDConnectProxy{})
+    httpcaddyfile.RegisterHandlerDirective("openid_connect_proxy", parseCaddyfile)
+}
```
noleto, line 134: Wouldn’t it be more flexible to handle the caddy registration in `module.go`? This would avoid registering it when the library is used for testing purposes, for example.
This approach would also align with the registration method used by other custom Caddy plugins.

lib/openidconnectproxy/auth.go
```diff
@@ -0,0 +1,139 @@
+package openidconnectproxy
```
noleto, line 1: It would be great to add some tests to this module.

natefikru: agreed

natefikru:
looks good, i will approve but would like to see a simple test file for this.

AusIV:
If this were something we intended to use for the long term I'd absolutely agree we should have tests. Given the tight timelines and temporary nature of this solution it felt unnecessary. As far as I'm concerned we can rip this code out as soon as we're ready to enable the service through #126

lib/openidconnectproxy/auth.go
AusIV: If this were intended for long term use I'd agree. Keeping as much as possible to one file will make pulling it back out easier once we integrate proper REST API support.

lib/openidconnectproxy/auth.go
AusIV: If we were keeping this around I'd absolutely agree. As this is a holdover solution until we have full REST support and then I intend to pull this code back out (especially knowing @natefikru has a solution integrated with the din middleware authentication in the works).

cds-amal: I think the team consensus is clear on this; where a test is required. I understand it's a band-aid, but even a band-aid should has a minimum `it works` test.

natefikru (approved):

PR #126, opened 2025-07-31
Feat/support bitcoin esplora

natefikru:
# Bitcoin Esplora Support & Caddyfile Parser Refactor

## Summary
This PR adds Bitcoin Esplora network support and refactors the Caddyfile parser for better maintainability.

### New Features
- **Bitcoin Esplora Handler**: Full support for Bitcoin blockchain via Esplora REST API
  - Block height queries
  - Transaction lookups
  - Address balance checks
  - Comprehensive test coverage

- **OAuth2 Authentication**: New OAuth2 client for provider authentication
  - Token management and refresh
  - Configurable refresh intervals

### Refactoring
- **Caddyfile Parser**: Extracted 400+ line `UnmarshalCaddyfile` function into dedicated `caddy_unmarshaller.go`
  - ~30 focused, single-responsibility functions
  - Consistent error handling
  - Reusable parsing helpers

### Fixes
- Fixed Beacon Chain request path parsing
- Improved error handling for REST API responses
- Fixed provider authentication initialization

## Files Changed
- **New**: `lib/network/bitcoin_esplora_handler.go`, `lib/auth/oauth2/client.go`, `modules/caddy_unmarshaller.go`
- **Modified**: Core middleware, network handlers, and provider authentication
- **Tests**: Added comprehensive test coverage for Bitcoin Esplora and OAuth2

## Testing
All existing tests pass. New test coverage added for:
- Bitcoin Esplora handler operations
- OAuth2 authentication flow
- Caddyfile parsing edge cases


docs/adding_new_networks.md
AusIV, line 1: I like the new form-based approach, but I think there's still utility in the old version for developer onboarding. Maybe make them separate files?

lib/auth/oauth2/client.go
```diff
@@ -0,0 +1,235 @@
...
+	ExpiresIn        int    `json:"expires_in"`
+	RefreshExpiresIn int    `json:"refresh_expires_in"`
+	Scope            string `json:"scope,omitempty"`
+	IDToken          string `json:"id_token,omitempty"`
+}
+
+// NewOAuth2Client creates a new OAuth2 client with the given configuration
+func NewOAuth2Client(config OAuth2Config) *OAuth2Client {
+	if config.RefreshIntervalSec == 0 {
+		config.RefreshIntervalSec = 240 // Default to 4 minutes
```
AusIV: I'd lean towards pulling this from the `expires_in` field on the request. Currently it's 5 mintues, but future providers may do something different, and it might even be variable (eg. expires at the top of the next hour).

modules/caddy_unmarshaller.go
AusIV, line 1: This is a huge improvement! This seems like an odd PR for it, but I'm happy to see it.

modules/din_middleware.go
```diff
@@ -349,8 +345,58 @@ func (d *DinMiddleware) startBackgroundServices() error {
...
 }
 
+// initializeOAuth2ForProvider initializes OAuth2 authentication for a provider
+func (d *DinMiddleware) initializeOAuth2ForProvider(provider *provider, networkObj *network, logger *logger.LoggerClient) error {
+	if networkObj.CustomConfig == nil {
+		return fmt.Errorf("OAuth2 enabled but no custom_config found")
+	}
+
+	// Extract OAuth2 configuration from custom_config
+	clientID, ok := networkObj.CustomConfig["oauth2_client_id"].(string)
```
AusIV: If we have more than one provider for an oauth authenticated network, this should be a provider level configuration option, not a network level one, as we'll have client ids and secrets on a per-provider basis.

modules/caddy_unmarshaller.go
natefikru: It's hard to not make some of these house keeping updates when working on this code. it was faster to just do it in the PR and its in a separate file so I just threw it in.

.gitignore
```diff
@@ -23,3 +23,8 @@ services/.DS_Store
 .DS_Store
 */.DS_Store
 coverage.out
+<<<<<<< HEAD
```
AusIV: This looks like git merge annotations got committed.

lib/auth/oidc/client.go
```diff
@@ -0,0 +1,328 @@
...
+			if c.logger != nil {
+				c.logger.Debug("using configured duration for token refresh",
+					zap.Int("duration_seconds", c.DurationSeconds),
+					zap.Duration("sleep_duration", sleepDuration))
+			}
+		} else {
+			// Use token expiry minus 1 minute (existing logic)
+			sleepDuration = time.Until(currentExpiry.Add(-1 * time.Minute))
+			if sleepDuration < 30*time.Second {
+				sleepDuration = 30 * time.Second
```
AusIV: This seems like it could cause problems if we're ever faced with an endpoint that has short token expirations. `currentExpiry` is provided by the target API - outside of our control.  If they gave us 20 seconds, `sleepDuration` would be `-40 * time.Second`, which is less than `30 * time.Second`, so we'd set `sleepDuration  = 30 *time.Second` and not start the token refresh until 10 seconds after the token has expired.

lib/openidconnectproxy/auth.go
AusIV, line 1: I'd be inclined to leave this in place until we've gotten Infura / Metamask to transition from the old endpoint to the new endpoint.  They're currently accessing through `esplora.rivet.link/esplora`. I'd like to be able to continue offering that while asking them to switch to `ingress.din.dev/bitcoin-esplora`, and then pull this code out for the next update after they've done that.

modules/provider.go
```diff
@@ -26,6 +27,12 @@ type provider struct {
 	// Registry Configuration Values
 	Methods map[string]struct{}  `json:"methods"`
 	Auth    *siwe.SIWEClientAuth `json:"auth"`
+	
+	// OIDC client for OAuth2/OIDC authentication
+	OIDCClient *oidc.OIDCClient `json:"oidc_client"`
```
AusIV, line 32: This works for now, but we should think about how to avoid having a separate struct field for every auth client we want to support. They all implement the same interface, but interfaces get lost between the UnmarshalCaddyfile phase and the Provision phase. I think maybe some kind of generalized auth config set up during the Unmarshal phase that can be used to instantiate an auth.IAuthClient during the provision phase. Again, this isn't a now problem, but I'd like to address it before we have a third auth client.

AusIV:
Another note: Having seen a sample size of more than one provider of the Bitcoin Esplora API, it appears the `/api` prefix is not a part of the API specification, but a part of Blockstream's implementation.

The Validation Cloud endpoint for blocks/tip/hash looks like `https://mainnet.bitcoin.validationcloud.io/v1/{KEY}/blocks/tip/hash`, so we shouldn't assume our target will always expect to receive the `/api` prefix

lib/auth/oidc/client.go
natefikru: updated

.gitignore
natefikru: updated

modules/provider.go
natefikru: agreed!

natefikru:
> Another note: Having seen a sample size of more than one provider of the Bitcoin Esplora API, it appears the `/api` prefix is not a part of the API specification, but a part of Blockstream's implementation.
> 
> The Validation Cloud endpoint for blocks/tip/hash looks like `https://mainnet.bitcoin.validationcloud.io/v1/{KEY}/blocks/tip/hash`, so we shouldn't assume our target will always expect to receive the `/api` prefix

addressed

AusIV (approved):

PR #132, opened 2025-08-19
Differentiate multiple provider endpoints from same service in metrics

natefikru:
- Extract API key suffix from URL path or headers to create unique provider
  identifiers
  - Handle collision detection when multiple providers have the same suffix
  - Retroactively update the first provider with a suffix when a second provider with
  the same base host is added
  - Comprehensive test coverage for all edge cases

  Example:
  Before: All 3 validation cloud endpoints → mainnet.bsc.validationcloud.io/{key-(suffix)}
  After:
  - Endpoint 1 → mainnet.bsc.validationcloud.io-1234 // last 4 chars of the api key
  - Endpoint 2 → mainnet.bsc.validationcloud.io-5678
  - Endpoint 3 → mainnet.bsc.validationcloud.io-9abc


AusIV (approved):

twhay:
This looks good to me - will be very helpful for understanding if the relative lags and jump in block numbers we are seeing in responses is coming from one of these endpoints.

PR #134, opened 2025-08-20
Dynamic Block Lag Limit Calculation

natefikru:
## Summary

Adds automatic calculation of block lag limits based on actual blockchain block production rates. Different blockchain networks produce blocks at varying speeds (e.g., Solana ~400ms, BSC ~3s, Starknet ~2s, Bitcoin ~10min), so a static block lag limit is suboptimal for provider health monitoring. This feature calculates the appropriate limit for each network on startup.

## How It Works

On network startup, the system:
1. Retrieves the latest block number from a healthy provider
2. Fetches timestamps for two blocks: the latest block and a block 10 blocks earlier
3. Calculates average block time from the timestamp difference
4. Computes the optimal block lag limit: `13000ms / avg_block_time_ms`
5. Rounds up to the nearest multiple of 5, with a minimum of 5 blocks

The calculation uses immutable blockchain timestamps, ensuring deterministic results across all server instances querying the same blocks.

## Supported Handlers

- **EVM** (Ethereum, BSC, zkSync, SEI, etc.)
- **Solana**
- **Starknet**
- **Bitcoin**

Handlers that don't support dynamic calculation (Beacon Chain, Bitcoin Esplora) retain their configured default limits.

## Interface Additions

Two methods added to `NetworkHandler`:
- `SupportsDynamicBlockLag() bool` - indicates whether the handler supports dynamic calculation
- `GetBlockTimestamp(blockNumber int64) (int64, error)` - retrieves Unix timestamp for a specific block

## Default Block Lag Limit

The default block lag limit (`DefaultBlockLagLimit`) is set to **15 blocks**. This conservative default provides:
- Adequate tolerance for networks where dynamic calculation fails or is not supported
- Safety margin for EVM networks with variable block times during network congestion
- Coverage for archive networks that may have higher provider sync variance

For most EVM networks (~2s block time), dynamic calculation yields 5-10 blocks. For faster chains like Solana (~400ms), it yields ~30-35 blocks. The 15-block default serves as a safe middle ground when dynamic calculation is unavailable.

## Fallback Behavior

If the calculation fails (no providers available, network errors, etc.), the system retains the configured default `BlockLagLimit` value.

---

## Notes

**Immutable Timestamp-Based Calculation**: This implementation uses blockchain block timestamps rather than real-time observation to calculate block production rates. Block timestamps are immutable data baked into the blockchain itself, meaning any server querying blocks N and N-10 will always get the same timestamps and therefore compute the same block lag limit. This ensures deterministic, consistent results across all server instances without requiring coordination or shared state.


modules/consts.go
```diff
@@ -45,6 +45,8 @@ const (
 	DefaultHCTimeout               = 5
 	DefaultHCInterval              = 5
 	DefaultBlockLagLimit           = int64(5)
+	DefaultBlockLagPeriodMs        = 13000 // 13 seconds in milliseconds for dynamic block lag calculation
+	DefaultBlockMeasurementSeconds = 10    // Duration in seconds to measure block production rate
```
AusIV: What are the ramifications of having this be 10 on a network that has a block time of, say, 12 seconds? That's going to have between zero and one samples.

modules/network_block_lag.go
```diff
@@ -0,0 +1,182 @@
...
+	successfulMeasurements := 0
+	providerRates := make(map[string]float64)
+
+	for result := range results {
+		if result.err == nil && result.msPerBlock > 0 {
+			successfulMeasurements++
+			providerRates[result.name] = result.msPerBlock
+			if result.msPerBlock < fastestRateMs {
+				fastestRateMs = result.msPerBlock
+				fastestProvider = result.name
```
AusIV: If I'm reading this right, fastestProvider could be one that had fallen behind and happened to be catching up during the sampling window. That could significantly skew results on on networks with slower block times.

modules/network.go
```diff
@@ -266,17 +270,23 @@ func (n *network) evaluateProviderHealth(provider *provider, currentBlock int64,
 	// Check for block lag
 	var isLagged bool
 	var blockLag int64
+	
+	// Get block lag limit with mutex protection (used in multiple places below)
+	n.blockLagLimitMu.RLock()
+	blockLagLimit := n.BlockLagLimit
```
AusIV: I'd probably use [sync/atomic](https://pkg.go.dev/sync/atomic#LoadInt64)'s StoreInt64 / LoadInt64 rather than designating a mutex for this.

AusIV (commented):
I have some reservations with this. You can see in-line for a couple of questions, but doing this calculation dynamically seems like a good way to get discrepancies between different proxy instances in terms of what is considered healthy and what isn't.

If one server comes online at a time where a validator misses a block, they're going to have a different block lag limit than a server that comes online when all blocks were produced in a timely fashion. That means those two servers are going to have different block lag limits, and identify servers as unhealthy at different times. That would result in consistency issues as servers don't agree on what is the healthy set of nodes.

modules/network_block_lag.go
```diff
@@ -0,0 +1,184 @@
...
+	"sync/atomic"
+
+	"go.uber.org/zap"
+)
+
+// Block lag calculation constants
+const (
+	// BlockLagCalculationLookback is the number of blocks to look back when calculating block time
+	// Using 10 blocks provides a good balance between accuracy and speed
+	BlockLagCalculationLookback = 10
```
AusIV: The way we're doing this now, we just query for the two values, we don't actually have to wait this many blocks, so I think I'd do 1000 (or 1024 because I like powers of two) to get much more tuned into the chain's block speed and have reorgs or other delays turn into less than a rounding error.

modules/network_block_lag.go
```diff
@@ -0,0 +1,184 @@
...
+			zap.Int64("default_limit", atomic.LoadInt64(&n.BlockLagLimit)))
+		return
+	}
+
+	// Try each provider until we get successful timestamps
+	var avgBlockTimeMs float64
+	var successfulProvider string
+
+	for providerName, p := range n.Providers {
+		blockTimeMs, err := n.measureBlockTimeFromTimestamps(p)
```
AusIV: I don't know how much value we get from doing this per provider anymore. If they don't agree with the provided block time stamps, we have bigger problems. Not a showstopper, and not horribly inefficient since I understand this is done once at startup, but it seems excessive.

lib/network/json_rpc_client.go
```diff
@@ -191,8 +191,15 @@ func PerformGetBlockByNumberViaJSONRPC(httpUrl string, headers map[string]string
 	var getBlockMethod string
 	for _, method := range supportedMethods {
 		methodLower := strings.ToLower(method)
+		// Match EVM-style "eth_getBlockByNumber", Starknet-style "starknet_getBlockWithTxs", or Solana-style "getBlock"
 		if strings.Contains(methodLower, "getblockbynumber") ||
-			strings.Contains(methodLower, "get_block_by_number") {
+			strings.Contains(methodLower, "get_block_by_number") ||
```
AusIV: I feel like this could get pretty cumbersome for something that could be known to the handler in advance. Not a blocker, but something we should consider in the future.

natefikru:
Superseded by #176 with a clean branch and all PR feedback addressed (1024 block lookback, simplified provider selection, direct method passing).

PR #137, opened 2025-08-28
feat(network): add guards to NetworkHandler implementations

smoyer64:
A simple PR to cause compiler errors if:

1. The `NetworkHandler` interface is changed in a way that causes existing handlers to no longer be implementations.
2. An implementation is changed in a way that causes it to no longer conform to the `NetworkHandler` interface.


natefikru:
nice defensive programming.

natefikru (approved):

PR #142, opened 2025-09-11
feat(tron): implement the Tron Full Node HTTP API

smoyer64:
This PR implements a DIN `network.Handler` for the [Tron Full Node HTTP API](https://developers.tron.network/reference/full-node-api-overview).  The action happens in `lib/network/tron_handler.go` with the following notable points:

* Retry logic is handled by the [Backoff module](https://pkg.go.dev/github.com/cenkalti/backoff/v5)
* Client errors (except for rate-limits) are _not_ retried.
* `IsRetryableError` could be made more efficient if the identified network errors are checked using `errors.Is()`

Add this handler to the `Caddyfile` using the following snippet (notice that the `chainstack` provider URL is the same as that for the JSON-RPC except the `/jsonrpc` path segment has been removed):

```Caddyfile
tron-nile-full {
	handler tron-full-node
	providers {
		https://tron-nile.core.chainstack.com/<api key> {
			priority 0
		}
	}
	archive_enabled false
	chain_id 0xcd8690dc
}		}

```

Resolves https://www.notion.so/consensys/DIN-Router-HTTP-Handler-for-Tron-Network-25cfc61a326e80849484c8d6c5710e4c


lib/network/tron_handler.go
```diff
@@ -0,0 +1,657 @@
...
+}
+
+func (h *TronHandler) performGetBlockByNumber(httpUrl string, headers map[string]string, httpClient din_http.IHTTPClient, authClient auth.IAuthClient, blockNumber int64) func() (*TronBlock, error) {
+	return func() (*TronBlock, error) {
+		reqBody, err := h.CreateBlockRequest("", blockNumber, false)
+		if err != nil {
+			return nil, err
+		}
+
+		respBody, statusCode, err := httpClient.Post(httpUrl+"/wallet/getblock", headers, reqBody, authClient)
```
natefikru: also use defined var/getMethod function here too

lib/network/tron_handler.go
```diff
@@ -0,0 +1,657 @@
...
+}
+
+func (h *TronHandler) getChainID(httpUrl string, headers map[string]string, httpClient din_http.IHTTPClient, authClient auth.IAuthClient) func() (string, error) {
+	return func() (string, error) {
+		reqBody, err := h.CreateBlockRequest("", 0, false)
+		if err != nil {
+			return "", err
+		}
+
+		respBody, statusCode, err := httpClient.Post(httpUrl+"/wallet/getblock", headers, reqBody, authClient)
```
natefikru: also use defined var/getMethod function here too

lib/network/tron_handler.go
```diff
@@ -0,0 +1,657 @@
...
+		context.Background(),
+		h.getChainID(httpUrl, headers, httpClient, authClient),
+		backoff.WithBackOff(backoff.NewExponentialBackOff()),
+		backoff.WithMaxTries(uint(requestAttempts)),
+	)
+}
+
+func (h *TronHandler) getChainID(httpUrl string, headers map[string]string, httpClient din_http.IHTTPClient, authClient auth.IAuthClient) func() (string, error) {
+	return func() (string, error) {
+		reqBody, err := h.CreateBlockRequest("", 0, false)
```
natefikru, line 569: so the "chain id" in this case is just the genesis block hash?

lib/network/tron_handler.go
```diff
@@ -0,0 +1,657 @@
...
+	TXTrieRoot       string `json:"txTrieRoot"`
+	ParentHash       string `json:"parentHash"`
+	Number           int64  `json:"number"`
+	WitnessID        int64  `json:"witness_id"`
+	WitnessAddress   string `json:"witness_address"`
+	Version          int64  `json:"version"`
+	AccountStateRoot string `json:"accountStateRoot"`
+}
+
+type TronTransaction struct {
```
natefikru: should this be just removed?

lib/network/tron_handler.go
```diff
@@ -0,0 +1,657 @@
...
+	block, err := h.parseBlockResponse(body)
+	if err != nil {
+		return "", err
+	}
+
+	if !block.IsGenesis() {
+		return "", fmt.Errorf("%w: block number %d", ErrNotTronGenesisBlock, block.BlockHeader.RawData.Number)
+	}
+
+	return "0x" + block.BlockID[len(block.BlockID)-8:], nil
```
natefikru, line 544: lets leave a comment here explaining the logic behind this return

lib/network/tron_handler.go
```diff
@@ -0,0 +1,657 @@
...
+func (h *TronHandler) GetLatestBlockNumber(httpUrl string, headers map[string]string, httpClient din_http.IHTTPClient, authClient auth.IAuthClient, requestAttempts int) (*LatestBlockResult, error) {
+	return backoff.Retry(
+		context.Background(),
+		h.getLatestBlockNumberFunc(httpUrl, headers, httpClient, authClient),
+		backoff.WithBackOff(backoff.NewExponentialBackOff()),
+		backoff.WithMaxTries(uint(requestAttempts)),
+	)
+}
+
+func (h *TronHandler) getLatestBlockNumberFunc(httpUrl string, headers map[string]string, httpClient din_http.IHTTPClient, authClient auth.IAuthClient) func() (*LatestBlockResult, error) {
```
natefikru: is there a reason `getLatestBlockNumberFunc` is separated from the `GetLatestBlockNumber` code block other than readability?

lib/network/tron_handler.go
```diff
@@ -0,0 +1,657 @@
...
+
+// ConfigureRequestPath implements the Handler interface.
+func (h *TronHandler) ConfigureRequestPath(req *http.Request, providerPath string, networkName string) error {
+	ConfigureRESTRequestPath(req, providerPath, networkName)
+
+	return nil
+}
+
+// ParseResponse implements the Handler interface.
+func (h *TronHandler) ParseResponse(body []byte, statusCode int) error {
```
natefikru, line 151: we aren't checking for errors in the potential 200 responses here. Have you double checked to make sure that this would never happen? Look at beacon_handler.go for reference

lib/network/tron_handler.go
```diff
@@ -0,0 +1,657 @@
...
+
+	if !block.IsGenesis() {
+		return "", fmt.Errorf("%w: block number %d", ErrNotTronGenesisBlock, block.BlockHeader.RawData.Number)
+	}
+
+	return "0x" + block.BlockID[len(block.BlockID)-8:], nil
+}
+
+// ValidateChainID implements the Handler interface.
+func (h *TronHandler) ValidateChainID(chainID string) error {
```
natefikru, line 548: we shouldn't be doing the actual equal check here, instead we check to see that the chain id is in present and in the correct format. beacon_handler.go for reference

lib/network/tron_handler.go
```diff
@@ -0,0 +1,657 @@
...
+// Block operations
+//
+
+// FormatBlockHeight implements the Handler interface.
+func (h *TronHandler) FormatBlockHeight(blockNum int64) string {
+	return fmt.Sprintf("%d", blockNum) // Decimal format for Tron
+}
+
+// CreateBlockRequest implements the Handler interface.
+func (h *TronHandler) CreateBlockRequest(_ string, blockNum int64, includeTransactions bool) ([]byte, error) {
```
natefikru, line 224: just double checking, is it possible to send onchain transactions via tron full node?

lib/network/tron_handler.go
```diff
@@ -0,0 +1,657 @@
...
+// ParseArchiveCheck implements the Handler interface
+func (h *TronHandler) PerformArchiveCheck(_ string, _ map[string]string, _ din_http.IHTTPClient, _ auth.IAuthClient, _ int, _ string) error {
+	return fmt.Errorf("%w: the TronHandler does not support PerformArchiveCheck", ErrUnsupportedFeature)
+}
+
+//
+//
+//
+
+func (h *TronHandler) parseBlockResponse(body []byte) (*TronBlock, error) {
```
natefikru, line 629: is there a reason why this is separate from ParseBlockResponse?

lib/network/tron_handler.go
```diff
@@ -0,0 +1,657 @@
...
+			Detail  bool   `json:"detail"`
+		}{
+			IDOrNum: h.FormatBlockHeight(blockNum),
+			Detail:  includeTransactions,
+		},
+	)
+}
+
+// ParseBlockResponse implements the Handler interface.
+func (h *TronHandler) ParseBlockResponse(body []byte) (interface{}, error) {
```
natefikru, line 237: This function wrapping seems unneeded logically.

lib/network/tron_handler.go
```diff
@@ -0,0 +1,657 @@
...
+		// Pending Pool - https://developers.tron.network/reference/gettransactionlistfrompending (3)
+		"/wallet/gettransactionlistfrompending",
+		"/wallet/gettransactionfrompending",
+		"/wallet/getpendingsize",
+	}
+}
+
+// GetBlockByNumberMethod implements the Handler interface.
+func (h *TronHandler) GetBlockByNumberMethod() string {
+	return tronBlockInfoEndpoint
```
natefikru, line 360: I think this may be incorrect, right now it is pointing to /wallets/getnowblock which i believe is the method for getting the latest block, not getting a block by number/hex

lib/network/tron_handler.go
```diff
@@ -0,0 +1,657 @@
...
+		},
+	}, nil
+}
+
+//
+// ChainID operations
+//
+
+// GetChainIDMethod implements the Handler interface.
+func (h *TronHandler) GetChainIDMethod() string {
```
natefikru, line 519: This isn't being used anywhere

lib/network/tron_handler.go
```diff
@@ -0,0 +1,657 @@
...
+
+	return block.BlockHeader.RawData.Number, nil
+}
+
+// PerformGetBlockByNumber implements the Handler interface.
+func (h *TronHandler) PerformGetBlockByNumber(httpUrl string, headers map[string]string, httpClient din_http.IHTTPClient, authClient auth.IAuthClient, requestAttempts int, blockNumber int64) (interface{}, error) {
+	return backoff.Retry(
+		context.Background(),
+		h.performGetBlockByNumber(httpUrl, headers, httpClient, authClient, blockNumber),
+		backoff.WithBackOff(backoff.NewExponentialBackOff()),
```
natefikru: I don't think we should be doing retries within the handler itself, this should be done at the application level. Otherwise there will be multiple levels of retries happening. I'm not opposed to this way of retrying but it should be done elsewhere.

lib/network/tron_handler.go
smoyer64: Nope ... and I've move those closures into the exported method which has the added advantage of the passed parameters being directly available (instead of being passed into a function to return the closure.)

lib/network/tron_handler.go
smoyer64: The inner `parseBlockResponse` is also used by `GetLatestBlockNumber`, `PerformGetBlockByNumber`, `ParseBlockNumberResponse` and `ParseChainIDResponse` ... I do see that I haven't actually "used" some of those calls and will be fixing those.  Critically, `GetLatestBlockNumber` and `GetChainID` do further processing to the returned `TronBlock` which makes returning `interface{}` inconvenient.

lib/network/tron_handler.go
smoyer64: Added code to make common network errors retryable and note to describe additional error information that might be returned in the HTTP response body.

natefikru: we still aren't checking for application level errors like this.

Screenshot 2025-09-17 at 2 04 24 PM

lib/network/tron_handler.go
natefikru: I'm still unsure why we are doing this when we have retries happening at the higher level.

lib/network/tron_handler.go
natefikru: no comment found

smoyer64: Ugh ... sorry about that.  Apparently I dragged it to the wrong method when I decided that the description would be helpful as part of the Go docs.  Fixed (and checked and added a link)

lib/network/tron_handler.go
smoyer64: Had this as test code in an uncommitted file (did analysis for it twice) ... just committed in "the right place".

natefikru (approved):

PR #146, opened 2025-09-18
Refactor: Centralize retry logic and eliminate redundant retries

natefikru:
# Refactor: Centralize retry logic and eliminate redundant retries

**Base:** `feat/tron-full-node`
**Compare:** `feat/tron-full-node-efficient-retries`

## Summary
- Removes redundant retry logic from all network handlers (Level 2) to eliminate request multiplication
- Creates centralized retry mechanism in `lib/network/retry.go`
- Ensures retry logic exists only at middleware/higher level (Level 1)

## Problem Solved
Previously, both middleware and handlers were implementing retry logic, causing up to 9x request multiplication. This PR consolidates all retry logic to a single layer.

## Changes
- **Removed retry logic from handlers**: Tron, Beacon, Bitcoin, Bitcoin Esplora, and shared JSON-RPC helpers now make single attempts only
- **Added centralized retry functions**: `Retry()` and `RetryWithChecker()` in `lib/network/retry.go`
- **Updated middleware**: Uses new centralized retry functions for registry calls and network operations
- **Comprehensive tests**: Added full test coverage for retry logic including context cancellation and exponential backoff

### DIFF
--- a/lib/network/beacon_handler.go
+++ b/lib/network/beacon_handler.go
@@ -486,160 +486,150 @@ func (h *BeaconChainHandler) ParseChainIDResponse(body []byte, statusCode int) (
 // For beacon chain, we can either return the configured chain ID directly
 // or make a REST API call to get the network configuration
 func (h *BeaconChainHandler) GetChainID(httpUrl string, headers map[string]string, httpClient din_http.IHTTPClient, authClient auth.IAuthClient, requestAttempts int) (string, error) {
-	var lastErr error
-	for attempt := 0; attempt < requestAttempts; attempt++ {
-		// Make GET request to config/spec endpoint
-		configURL, err := url.JoinPath(httpUrl, h.GetChainIDMethod())
-		if err != nil {
-			lastErr = fmt.Errorf("failed to construct config URL: %w", err)
-			continue
-		}
-		resBytes, statusCode, err := httpClient.Get(configURL, headers, authClient)
-		if err != nil {
-			lastErr = fmt.Errorf("error sending HTTP request: %w", err)
-			continue
-		}
-
-		// Parse the response to extract chain ID
-		chainID, err := h.ParseChainIDResponse(resBytes, *statusCode)
-		if err != nil {
-			lastErr = err
-			continue
-		}
-
-		// Success!
-		return chainID, nil
+	// Single attempt - let middleware handle retries
+	configURL, err := url.JoinPath(httpUrl, h.GetChainIDMethod())
+	if err != nil {
+		return "", fmt.Errorf("failed to construct config URL: %w", err)
+	}
+	resBytes, statusCode, err := httpClient.Get(configURL, headers, authClient)
+	if err != nil {
+		return "", fmt.Errorf("error sending HTTP request: %w", err)
 	}
 
-	return "", fmt.Errorf("failed after %d attempts: %w", requestAttempts, lastErr)
+	// Parse the response to extract chain ID
+	chainID, err := h.ParseChainIDResponse(resBytes, *statusCode)
+	if err != nil {
+		return "", err
+	}
+
+	return chainID, nil
 }
 
 // GetLatestBlockNumber retrieves the latest block number (slot) for beacon chain
 // Uses the REST API endpoint /eth/v2/beacon/blocks/head to get the latest slot
 func (h *BeaconChainHandler) GetLatestBlockNumber(httpUrl string, headers map[string]string, httpClient din_http.IHTTPClient, authClient auth.IAuthClient, requestAttempts int) (*LatestBlockResult, error) {
-	var lastErr error
-	var lastResponseStatus int
-	var lastHealthStatus = Unhealthy
-
-	// Use the block info endpoint to get latest slot
+	// Single attempt - let middleware handle retries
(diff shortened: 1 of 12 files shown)


lib/network/retry.go
```diff
@@ -0,0 +1,90 @@
+package network
+
+import (
+	"context"
+	"time"
+)
+
+// Retry performs an operation with exponential backoff up to maxAttempts times.
+// It will retry on any error. Use RetryWithChecker for custom retry logic.
+func Retry(ctx context.Context, maxAttempts int, operation func() error) error {
```
smoyer64: Is there a reason not to use the `backoff` library to do the looping and exponential backoff here? (and `RetryWithChecker`?)

AusIV: Can you provide a link to the backoff library you're referring to? There's not one in the standard library, and I'm finding a couple of different backoff libraries on Google.

modules/network.go
```diff
@@ -441,7 +472,26 @@ func (n *network) performArchiveCheck(provider *provider, currentBlock int64) er
 	// Use handler method to format block height directly
 	quarterBlockHeightString := n.handler.FormatBlockHeight(quarterBlockHeight)
 
-	return n.handler.PerformArchiveCheck(provider.HttpUrl, provider.Headers, n.HttpClient, provider.AuthClient(), n.RequestAttemptCount, quarterBlockHeightString)
+	return networklib.RetryWithChecker(
+		context.Background(),
+		n.RequestAttemptCount,
+		func(err error) bool {
+			retryable := n.handler.IsRetryableError(err, 0)
+			if !retryable {
```
AusIV: In general, I think code like this would be cleaner if we did:

```
n.loggerDebug("Error in archive check",
   zap.Error(err),
   zap.Bool("retryable", retryable),
   zap.String("provider", provider.host))
```

The debug messages aren't quite as clean, but all the information is there, and the code gets a lot more concise.

I could probably be convinced otherwise for Info / Warning / Error level logs, but for debug level logs having concise code seems preferable to verbose error messages so long as all the information is available.

AusIV (commented):
Over all I think it's a good direction. I do like @smoyer64's idea of using libraries that specialize in retries, especially given the benefits we can get from backoff policies.

lib/network/beacon_handler.go
```diff
@@ -760,3 +750,11 @@ func getBoolFromMetadata(metadata map[string]interface{}, key string) bool {
...
 	return false
 }
+
+// Helper function to get minimum of two integers
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
```
noleto: Couldn't we just use the standard `math.Min(a, b)`?

lib/network/retry.go
noleto: In the Python world, there’s a great Retry strategy package, but I couldn’t find an equivalent in Golang. If we stick to our own implementation, it’s worth borrowing ideas to make the Retry logic more flexible. For reference, check out:  
- https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry  
- https://medium.com/@hudsonbrendon/using-retry-in-http-requests-with-python-5c46e3280893

lib/network/beacon_handler.go
natefikru: done

lib/network/retry.go
```diff
@@ -0,0 +1,242 @@
...
+				zap.Duration("attempt_duration", attemptDuration),
+				zap.Duration("total_elapsed", time.Since(startTime)),
+			)
+
+			if !retryable {
+				logger.Debug("Non-retryable error encountered", logFields...)
+				return nil, backoff.Permanent(err)
+			}
+
+			if attemptCount < maxAttempts {
```
smoyer64, line 177: You don't need to keep track of whether to retry or not - just return the `err` and `backoff` will retry (it's the default.)  You've properly wrapped the `err` returned in the block above this one in `backoff.Permanent(err)` which will cause `backoff` to stop retrying. (2x)

lib/network/retry.go
```diff
@@ -0,0 +1,242 @@
...
+	maxAttempts int,
+	isRetryable func(error) bool,
+	operation func() error,
+	fields ...zap.Field,
+) error {
+	attemptCount := 0
+	startTime := time.Now()
+
+	// Create exponential backoff
+	b := backoff.NewExponentialBackOff()
```
smoyer64, line 138: I don't think this is used anywhere - instantiate it when `backoff.Retry` is executed? (2x)

lib/network/retry.go
```diff
@@ -0,0 +1,242 @@
...
+				zap.Int("attempt", attemptCount),
+				zap.Duration("attempt_duration", attemptDuration),
+				zap.Duration("total_duration", time.Since(startTime)),
+			)...,
+		)
+		return nil, nil
+	}
+
+	// Create notify function for retry attempts
+	notifyFunc := func(err error, duration time.Duration) {
```
smoyer64, line 197: Is this going to result in duplicate attributes since (some of) these are also set in L165-L169?

smoyer64 (changes requested):

PR #147, opened 2025-09-18
Add API Key Metric for requests

AusIV:
We're going to start requiring keys for access to the gateway, and we want to know which keys are making which requests.

This adds the api key to prometheus metrics, and passes it through from the request.

### DIFF
--- a/lib/prometheus/prometheus.go
+++ b/lib/prometheus/prometheus.go
@@ -61,7 +61,7 @@ func RegisterMetrics() {
 			Name: DinRequestCountMetricName,
 			Help: "Metric for counting the number of requests to the din http server",
 		},
-		[]string{"service", "method", "provider", "host_name", "response_status", "health_status", "priority", "machine_id", "environment"},
+		[]string{"service", "method", "provider", "api_key", "host_name", "response_status", "health_status", "priority", "machine_id", "environment"},
 	)
 	DinRequestDurationMilliseconds = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
@@ -133,6 +133,7 @@ type PromRequestMetricData struct {
 	Method         string
 	Network        string
 	Provider       string
+	ApiKey         string
 	HostName       string
 	ResponseStatus int
 	HealthStatus   string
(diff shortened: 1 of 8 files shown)


smoyer64 (changes requested):
The incoming `ApiKey` is a secret.  Instead of pushing that out into our observability DB, we should look up the identity of the user and tag with that instead.  I'm assuming the code to actually block the request isn't included since we need to wait until all users are sending their `ApiKey`.  Once we want to block invalid header values, we'll need a set containing those that are allowed.  We can use a map instead and the lookup can happen in the same step.  Want to add that code and comment out the blocking part?

AusIV:
Hmmm... You're right. I hadn't thought about the fact that SigNoz is accessible to people we wouldn't want to have the raw API keys. 

At this point we're not implementing blocking because we want visibility to who is providing the API key and who isn't, but also for the short term I was intending to implement blocking with off-the-shelf Caddy features, rather than building that into our middleware.

I'm also nervous at this stage to have to have a map configured while we're still trying to get adoption. If we find new teams that need keys, right now we can just make up a key for them, tell them to start using it, and start getting signoz data showing whether or not they are. If we have to have a map of keys to users, we'd have to update the map in the caddyfile and redeploy the router before we could see the data come into Signoz. That will become necessary later as we start actually requiring the keys, but right now that could be fairly frequent.

Let me ponder this problem for a bit.

AusIV:
@smoyer64's comment also made me realize that the header was being forwarded on to providers, and we probably don't want them to have it. I've got an update that strips the header before forwarding the request.

smoyer64:
> Hmmm... You're right. I hadn't thought about the fact that SigNoz is accessible to people we wouldn't want to have the raw API keys.

And for untrusted users, we don't want them to see (use) each other's tokens either.

noleto:
> > Hmmm... You're right. I hadn't thought about the fact that SigNoz is accessible to people we wouldn't want to have the raw API keys.
> 
> And for untrusted users, we don't want them to see (use) each other's tokens either.

Couldn't we hash the APIKey with SHA1, for instance, before pushing it to SigNoz? This way, we could still identify who is using an APIKey while keeping it secure.

AusIV:
So, I didn't see @noleto's suggestion until after I pushed my solution, but it's a mix of a map and hashing. If the key appears in the map, we use the value in the map. If the key isn't in the map, we submit a salted hash of it.

smoyer64:
> If the key isn't in the map, we submit a salted hash of it.

This is a great interim behavior but once we "require" a Consensys-generated key, those should be manually put into the map (and no other keys should be accepted.)  If another key _does_ come in, we should definitely log that!

noleto (approved):
Looks good to me

smoyer64:
Should we add a task to the backlog for the next step (denying unknown UUIDs)?

AusIV:
@smoyer64 I already have a separate ticket for that.

smoyer64 (approved):

PR #150, opened 2025-09-26
[Feature] Watcher Score and Dynamic Load Balancing

noleto:
This PR brings the dynamic routing feature to DIN Router as detailed in ticket #68. It is strongly recommended to read[ the design doc](https://github.com/DIN-center/din-caddy-plugins/blob/feature/68-watcher-score-dynamic-load-balancing/docs/dynamic-load-balancing/dynamic-load-balancing.md) before reviewing so that terminology and the overall workflow will be clear.

The PR is mainly reviewed in two parts, and following this order will make reviewing easier for you:
- Implementation of the watcher score module (in `lib/watcherscore`)
- Then, the dynamic load balancing workflow that leverages the reputation score (in `modules`)

The core logic was extensively tested using unit tests to ensure reproducibility (as we are using some randomness in the provider selection). See coverage report.

Manual integration tests are also possible based on the `Caddyfile.private` file that is not in the repo (You can find it in the team 1Passwaord vault - name `Caddyfile Dynamic Load Balacing test`)
- Cloning the repo (or updating the code to this branch feature)
- Build the docker image: `make docker-build `
- Launch the dev satck: `docker compose up -d`
- Trigger a few queries:

```shell
curl --location 'localhost:8000/bsc-mainnet' \
--header 'Content-Type: application/json' \
--header 'Din-Provider-Info: true' \
--data '{"method":"eth_blockNumber","params":[],"id":1,"jsonrpc":"2.0"}'
```

- For `eth`, there is only one provider, so 100% of requests are served by Rivet.  
- For `bsc-mainnet`, there are two providers, so the traffic will be distributed between them according to the Watcher scores.  
  You can check their scores in the logs at `<project_root>/caddy-data/caddy.log`.

Loggers are in Debug mode so you can follow the workflow. Loggers related to this feature start with:
- `[WATCHER_SCORE]...` for all loggers related to watcher score logic
- `[DYNAMIC_LB]...` for all loggers related to the DIN proxy logic that uses provider scores to dynamic load balance the traffic


lib/watcherscore/transformers.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+	}
+	return transformedScores, nil
+}
+
+func NewCompositeTransformer(chain ...ScoreTransformer) *CompositeTransformer {
+	return &CompositeTransformer{chain: chain}
+}
+
+// EWMATransformer is a transformer that applies an Exponential Weighted Moving Average (EWMA) to the scores.
```
AusIV, line 31: In its current implementation, I don't think we can use the EWMATransformer in production, because we couldn't ensure that different router instances have the same values when they didn't start up at the same time. If the different router instances score differently, then which provider your request goes to is dependent on which router instance picks it up. And while the scores should get closer together the longer the servers run, I don't think they'll ever actually reach the same routing decisions.

This might be addressable by having a shared state (redis?) for previous scores that could allow nodes just coming online to start with the same previousScores value, which I think should keep them in sync. There are several other angles arising for which a location for shared state is going to be necessary for upcoming features, so this might be worth addressing soon.

lib/watcherscore/watcherscore_manager.go
```diff
@@ -0,0 +1,223 @@
...
+	if _, networkExists := rm.scores[network]; !networkExists {
+		return NewEmptyScore()
+	}
+	if _, providerExists := rm.scores[network][providerID]; !providerExists {
+		return NewEmptyScore()
+	}
+	return rm.scores[network][providerID]
+}
+
+func (rm *WatcherScoreManager) StartPeriodicUpdates(frequency time.Duration) chan struct{} {
```
AusIV, line 154: Not being especially familiar with the watcher API - how frequently will the values returned by the watcher API update? If one server updates now and the next updates in a minute, will they get the same score? Are there boundaries where updates occur, or do the update continuously as the watcher gets new information?

I'm thinking it may make sense to try to synchronize the updates by starting the process at the nearest interval of the frequency. Eg. If the frequency is 5 minutes and `now % 300 seconds == 180 seconds`, maybe do the initial computation now, the next computation in 120 seconds, and then start polling at the 5 minute interval to make sure the different server instances are getting as close to the same data as possible.

modules/caddy_unmarshaller.go
```diff
@@ -609,6 +611,54 @@ func (p *caddyfileParser) parseDinRegistry() error {
...
+				return p.dispenser.Errf("Error while parsing dynamic_load_balancing.enabled: %v", err)
+			}
+			p.middleware.DynamicLoadBalacingEnabled = boolValue
+		case "watcher_endpoint":
+			p.dispenser.Next()
+			p.middleware.WatcherApiEndpoint = p.dispenser.Val()
+		case "watcher_api_key":
+			p.dispenser.Next()
+			p.middleware.WatcherApiKey = p.dispenser.Val()
+		case "sync_score_interval_secs":
```
AusIV, line 632: Do we want a default here? If not we should include this field in the subsequent validation step.

modules/din_middleware.go
```diff
@@ -882,6 +945,29 @@ func (d *DinMiddleware) startRegistrySync() {
...
 }
 
+// startWatcherScoreSync starts a background goroutine to sync the watcher scores to the middleware
+func (d *DinMiddleware) startWatcherScoreSync() chan struct{} {
+	syncQuit := make(chan struct{})
+
+	// Do immediate initial sync
+	d.SyncMiddlewareWithLatestScores()
+
+	go func() {
```
AusIV, line 959: Given my comment on trying to synchronize router instances on their polling interval, it probably makes sense to synchronize this to happen just after that step.

modules/din_select.go
```diff
@@ -37,8 +38,13 @@ func (DinSelect) CaddyModule() caddy.ModuleInfo {
 func (d *DinSelect) Provision(context caddy.Context) error {
 	d.logger = context.Logger(d)
 
-	selector := &reverseproxy.HeaderHashSelection{Field: "Din-Session-Id"}
-	if err := selector.Provision(context); err != nil {
+	selector := &reverseproxy.HeaderHashSelection{Field: "Din-Session-Id",
```
AusIV, line 41: If I'm understanding this correctly, the din score based selector won't see much use in practice, because the vast majority of the requests we get specify the Din-Session-Id.

I had been under the impression that we were going to use the hashed value of the Din-Session-Id to map to a provider weighted based on scores. We would expect a Din-Session-Id to map to the same provider as long as weights did not change, but that the weights would assign sessions to providers in much the same way they assign requests without a session ID to providers.

That said, this understanding of the current code alleviates my other concerns about consistency of weights across different server instances, because if we're not passing session IDs we don't have consistency expectations anyway.

AusIV (commented):
I think this PR is okay in its current form, but mainly because it won't effect requests that set the Din-Session-Id header. The vast majority of the requests we get provide a Din-Session-Id header, and this PR is set to keep the old behavior for those requests, so the impact of this PR would be minimal on production traffic.

Ultimately, I'd like to see an update where sessions are assigned to providers in a similar fashion to how this assigns requests to providers, but before I can be comfortable with that we need to take measures to make sure the weights across different server instances are aligned. I've provided comments throughout the PR on how I think we could achieve that alignment.

modules/caddy_unmarshaller.go
noleto: `watcher_api_key` is required if the feature is enabled, so we have a post validation step at the end of parsing function to chekc if both, API Key and API endpoint was filled in (otherwise triggers an error):
https://github.com/DIN-center/din-caddy-plugins/blob/54f3cf6092ab780e47dd9a12c0a13a38a9d2832e/modules/caddy_unmarshaller.go#L645-L646

lib/watcherscore/transformers.go
noleto: You are correct. In the current version, where the logic for Watcher Score computation is embedded into the DIN Router instance, every time a new router instance starts, it fetches the latest Watcher checks results. As of now, the Watcher API returns the past "X minutes" of aggregated data (currently we use 5 minutes for aggregation) starting at the exact time the API is called. The Watcher Scores logic currently calls the Watcher API every 1 minute.  

Thus, if a router instance (where the Watcher Scores process runs) calls the Watcher API at a slightly different time (even a few seconds of difference), it is highly likely it won’t retrieve the exact same check results, consequently leading to slightly different scores.  

As you pointed out, in the long run, the scores will converge and get closer and closer, but they may never be exactly the same. We have a feature under discussion to implement on the Watcher API side called "deterministic outputs," which would alleviate this issue significantly (or possibly eliminate it).  

That said, this is probably not a major issue since the score is only used when no session affinity is requested by the user. When there is no session affinity, the scores determine the proportion of traffic each provider receives. Since the router operates across several different instances, this traffic distribution may differ slightly across instances, but it is unlikely to create a significant difference for providers with good standing scores.  

For providers with highly fluctuating metrics (e.g., frequent up-and-downs), the traffic share could vary slightly more. However, as the process runs over time, this difference will be adjusted and caught up, thanks to the behavior of the `EWMATransformer`.

lib/watcherscore/watcherscore_manager.go
noleto: > how frequently will the values returned by the watcher API update?

The Watcher API data updates every 1-minute, operating on a rolling 1-minute window. Every time you call the Watcher API, it returns aggregated data (the minimum aggregation is 1-minute, but we currently use a 5-minute aggregation in the Watcher Scores logic) for the latest X minutes, starting at **the exact moment** the API was called.  

By the way, this is something we plan to change soon: a new feature will return data for the most recent **"completed" minutes** instead of starting at the exact time of the call.

>  If one server updates now and the next updates in a minute, will they get the same score? 

Probably not. But in the long run, thanks to the `EWMATransformer`, the scores will catch up and get closer across all router instances, though likely not exactly the same.

> Are there boundaries where updates occur, or do the update continuously as the watcher gets new information?

Good point. As of now, Watcher checks are updated continuously as the Watcher gathers new information from the endpoints. When calling the Watcher API with a 5-minute aggregation level:  

- **Current behavior**:  
  - Call time: `14:38:20`  
  - It returns data for the interval: `14:33:20` (included) to `14:38:20` (included)
  - This can include incomplete data for the first or last 1-minute interval.  

- **Planned improvement**:  
  - Call time: `14:38:20`  
  - It should return data only for the interval: `14:33:00` (included) to `14:38:00` (excluded).  
  - This will ensure the data reflects **5 full, completed minutes**.  

The key benefit is that DIN Router instances, which call the Watcher API every 1 minute, will get consistent and deterministic outputs no matter the exact second the call is made. This will ensure all instances generate the exact same Watcher scores.

modules/din_select.go
noleto: This is correct. Session affinity (`Din-Session-Id`) takes precedence if provided by the user.  

When no session affinity is required, the weights (Watcher scores) determine the traffic share allocated to each provider. A key point to note is that the weights are updated from the Watcher Score module at a fixed interval. Currently, this is set to 90 seconds (configured using the `sync_score_interval_secs` option in the Caddyfile).  

This means that regardless of how frequently Watcher scores are updated, the scores (or weights) are synced to the DIN Router at a fixed time window. This ensures the traffic share remains stable and does not switch too frequently.

modules/din_middleware.go
```diff
@@ -107,12 +109,37 @@ type DinMiddleware struct {
...
+	watcherScoreManager ws.IWatcherScoreManager
+
+	//The watcher client for dynamic load balacing
+	watcherClient watcher.IWatcherAPIClient
+
+	// The channel to quit the goroutine that computes the watcher scores
+	watcherScoreComputeQuit chan struct{}
+
+	// The channel to quit the goroutine that syncs the watcher scores to the middleware
+	watcherScoreSyncQuit chan struct{}
```
natefikru: because there are so many watcher specific values i think it should be in its own object to keep this section a little bit more clean.

modules/din_middleware.go
```diff
@@ -163,6 +190,24 @@ func (d *DinMiddleware) initialize(context caddy.Context) error {
...
 
+	// If dynamic load balacing is enabled, initialize the score backend
+	if d.DynamicLoadBalacingEnabled {
+		d.logger.Info("[DYNAMIC_LB] Dynamic load balacing activated, initializing watcher score manager")
+		d.logger.Debug("[DYNAMIC_LB] Dynamic load balacing settings:",
+			zap.Uint64("watcher_scores_sync_interval_secs", d.WatcherScoreSyncIntervalSec),
+			zap.String("watcher_endpoint", d.WatcherApiEndpoint))
+
+		//list of networks to compute scores
+		networks := make([]string, 0, len(d.Networks))
```
natefikru, line 205: making a copy of the network list at start up makes it so that if a new network is added by the registry, it won't get picked up because the copy of networks was already created on initialization. ideally the list of networks is reinitialized on each sync interval.

lib/watcherscore/watcher_helpers.go
```diff
@@ -0,0 +1,161 @@
...
+	} else {
+		// Normalized exponential function with k=1
+		// f(x) = (e^(kx) - 1) / (e^k - 1)
+		normalizedLatency := (latencyStats.P95 - LowLatencyInMillis) / (HighLatencyInMillis - LowLatencyInMillis)
+		k := 1.0 // This is the slope of the curve, TODO: make this configurable
+		// We invert it since we want higher latencies to have lower scores
+		rank = 1.0 - (math.Exp(k*normalizedLatency)-1.0)/(math.Exp(k)-1.0)
+		// Visualization of the latency ranking function (k=1):
+		// Rank
+		// 1.0 |  ****
```
natefikru, line 48: This is awesome

lib/watcherscore/watcherscore_manager.go
```diff
@@ -0,0 +1,224 @@
...
+	}()
+
+	return stopChan
+}
+
+func (rm *WatcherScoreManager) GetAllScores(network string) map[string]*Score {
+
+	// Copy the scores to a new map to avoid returning the underlying map
+	scores := make(map[string]*Score)
+	for providerID, score := range rm.scores[network] {
```
natefikru, line 186: it looks like we have GetScore() + GetAllScores() reading from rm.scores map without a lock and we have ComputeScores() writing to it with a lock. so there is a potential race concern there. we should just add the read locks to the get functions.

modules/din_middleware_helpers.go
```diff
@@ -774,4 +788,45 @@ func (d *DinMiddleware) getAPIKeyId(key string) string {
...
+}
+
+// Fetches the latest score from the watcher score manager and updates the provider score for all active networks
+func (d *DinMiddleware) SyncMiddlewareWithLatestScores() {
+	// Lock the middleware object to prevent race condition when updating provider scores
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+
+	// Keep track of the last time the scores were synced to the middleware
+	d.WatcherScoreLastSyncTime = time.Now().UTC()
```
natefikru: we are modifying data while holding on to a read lock, this is a violation of how this type of lock should be used.

modules/din_middleware.go
```diff
@@ -107,12 +109,37 @@ type DinMiddleware struct {
...
+
 	// Map for associating API keys with users
 	ApiKeys map[string]string
 	ApiSalt string
 
-	// The channel to quit the goroutines
-	quit chan struct{}
+	// Dynamic load balacing configuration
+	// The flag to enable or disable the dynamic load balacing
+	DynamicLoadBalacingEnabled bool
```
natefikru: This is a typo, should be DynamicLoadBalancingEnabled

lib/watcherscore/watcherscore_manager.go
```diff
@@ -0,0 +1,224 @@
...
+		return NewEmptyScore()
+	}
+	return rm.scores[network][providerID]
+}
+
+func (rm *WatcherScoreManager) StartPeriodicUpdates(frequency time.Duration) chan struct{} {
+	stopChan := make(chan struct{})
+
+	// Do immediate initial update
+	rm.ComputeScores()
```
natefikru, line 158: for rm.ComputeScores() here and line 144, the error value is ignored, we should at least be logging.

modules/din_scorebased_selector.go
```diff
@@ -0,0 +1,152 @@
...
+
+					if providerScore.HasValue() {
+						s.logger.Debug("[DYNAMIC_LB] Score found for provider",
+							zap.String("provider", providerHost),
+							zap.Any("score", providerScore))
+
+						// If the score has a valid value and is not stale, we can use it to weight the selection
+						graceTimeStart := time.Now().UTC().Add(-time.Minute * StaleScoreGracePeriodInMinutes)
+						if provider.Score.LastUpdated().After(graceTimeStart) {
+							weight = int(provider.Score.Value() * 100)
```
natefikru: this 100 value should be named as a const. scoreToWeightMultiplier or something

modules/din_scorebased_selector.go
```diff
@@ -0,0 +1,152 @@
...
+		if v, ok := repl.Get(DinUpstreamsContextKey); ok {
+			providers = v.(map[string]*provider)
+		}
+
+		// If providers are available, we can select an upstream based on the watcher score
+		if providers != nil {
+			s.logger.Debug("[DYNAMIC_LB] Selecting upstream according to its watcher score")
+
+			// Create a map of upstreams and associated providers
+			mapUpstreamProviders := make(map[*reverseproxy.Upstream]*provider)
```
natefikru, line 89: we're looking up providers based on pointers, this is a little scary because if the mem address changes for some reason this could break, we should instead look up based on hostname

modules/din_scorebased_selector.go
```diff
@@ -0,0 +1,152 @@
...
+
+					providerHost = provider.host
+					providerScore = provider.Score
+
+					if providerScore.HasValue() {
+						s.logger.Debug("[DYNAMIC_LB] Score found for provider",
+							zap.String("provider", providerHost),
+							zap.Any("score", providerScore))
+
+						// If the score has a valid value and is not stale, we can use it to weight the selection
```
natefikru: the only concern here is that if the provider has a score of .9 for the first 60 minutes but is stale, then we automatically move them to .5. this would bring down that provider 40 percent and we are punishing them because our syncing is broken, instead is there a way to handle this gradually so at least it is not as abrupt and it would give us time to fix the issue?

This also applys for providers who are low at .1, we woudln't want to put the good providers and bad providers in the same weight range immeidately because of a syncing error. This does make sense though for providers that don't have any data to begin with. its only an issue for providers that have established some sort of score

modules/din_middleware.go
noleto: This is just the initial list (from the declared networks in the Caddyfile). If the registry is enabled, another part of the code handles the addition and removal of networks:
https://github.com/DIN-center/din-caddy-plugins/blob/13d635e7e8e665cd53d1330edc3f2cfecc26431d/modules/din_middleware_helpers.go#L212-L216
https://github.com/DIN-center/din-caddy-plugins/blob/13d635e7e8e665cd53d1330edc3f2cfecc26431d/modules/din_middleware_helpers.go#L135-L139

lib/watcherscore/watcherscore_manager.go
noleto: This logic code was updated indeed. The `ComputeScores()` function is now safe and error-free as all error handling is managed internally. It no longer returns errors.

modules/din_middleware.go
noleto: good catch, fixed

modules/din_middleware_helpers.go
noleto: indeed, changed to `d.mu.Lock()` (locks to writing)

lib/watcherscore/watcherscore_manager.go
noleto: well seen, added `rm.mu.RLock()` in both read methods

modules/din_middleware.go
noleto: Agree, too much conf in the same middleware. Created another object to wrap all config related to this feature.

modules/din_scorebased_selector.go
noleto: Done

modules/din_scorebased_selector.go
noleto: This mapping is created for every single request, and the address remains stable throughout its lifetime. This is the fastest way to map a Upstream to a provider, considering the frequency of calls. On a side note, addresses are ideal for map keys since they are uniformly distributed and hashable.

modules/din_scorebased_selector.go
noleto: Indeed, this is a good point. We probably need to adjust the score with an exponential function to "pull TO the mean" in case the score is stale (so it gradually converges to 50, the mean score). Let me give it a try...

noleto: Implemented an exponential function to gradually pull the scores toward the default value based on the elapsed time since the score became stale (refer to the plot to see how it behaves).
output

lib/watcherscore/consts.go
```diff
@@ -0,0 +1,15 @@
+package watcherscore
+
+const (
```
smoyer64, line 3: Ignoring the `XxxMetricID` constants, would we want any of these values to be configured via the Caddyfile?

lib/watcherscore/transformers_test.go
smoyer64, line 1: Should this file be `transformers_test.go`?

lib/watcherscore/transformation_test.go
```diff
@@ -0,0 +1,234 @@
...
+			"provider1": MustCreateScore(1.0, TIME1),
+			"provider2": MustCreateScore(0.0, TIME1),
+		}
+		scoresT := map[string]*Score{
+			"provider1": MustCreateScore(0.9, TIME1),
+			"provider2": MustCreateScore(0.1, TIME1),
+		}
+
+		// First call initializes previous scores
+		transformer.TransformScore("network", scoresTMinusOne)
```
smoyer64: This returns a result and an error but the error is not checked and the value isn't used.  I see _why_ that's done, but you could make this explicit which helps both humans and linters (6 typ):

```
_, _ = transformer.TransformScore("network", scoresTMinusOne)
```

lib/watcherscore/types.go
```diff
@@ -0,0 +1,152 @@
...
+	"math"
+	"net/url"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+// A provider metric is a number between 0 and 1 that can be used to measure the quality of a provider for a given criteria.
+// For example, the block number consistency metric ensures the consistency rate of a provider's block number.
+type ProviderMetric struct {
```
smoyer64, line 14: Nice and very much in the functional style! (I prefer immutable types anywhere but sometimes it's too painful)

lib/watcherscore/watcherscore_manager.go
```diff
@@ -0,0 +1,248 @@
...
+
+	// Make sure we only have unique networks
+	uniqueNetworksMap := make(map[string]struct{})
+	for _, network := range networks {
+		uniqueNetworksMap[network] = struct{}{}
+	}
+
+	rm := NewEmpty(logger)
+	for network := range uniqueNetworksMap {
+		rm.AddNetworkWithBuiltInFormula(network, client)
```
smoyer64: If this call returns an error, can we proceed by returning `rm` with one (or more) fewer networks?

modules/caddy_unmarshaller.go
```diff
@@ -609,6 +611,54 @@ func (p *caddyfileParser) parseDinRegistry() error {
...
+			p.dispenser.Next()
+			p.middleware.DynamicLoadBalancing.WatcherApiKey = p.dispenser.Val()
+		case "sync_score_interval_secs":
+			p.dispenser.Next()
+			syncScoreIntervalSecs := p.dispenser.Val()
+			intValue, err := strconv.Atoi(syncScoreIntervalSecs)
+			if err != nil {
+				return p.dispenser.Errf("Error parsing dynamic_load_balancing.sync_score_interval_secs: %v", err)
+			}
+			p.middleware.DynamicLoadBalancing.WatcherScoreSyncIntervalSec = uint64(intValue)
```
smoyer64: This could be an integer overflow - though I hope the watcher sync interval is never that long!  This and other sync intervals are used as `time.Durations`, so should we instead just initialize them as `int64` to begin with?  (and take the `abs()` to make sure they're positive?)

docs/dynamic-load-balancing/dynamic-load-balancing.md
smoyer64, line 1: This is a great introduction document!  As a relative newb to the router, I can honestly say it made a huge difference.  One question I have is about the diagrams - do we tend to eschew the use of Mermaid mark-up on the DIN team?  A simple example is the SoC diagram which could be:

```mermaid
graph TD;
    subgraph WatcherScoreManager
        direction TB
        subgraph ScoreFormula
            C1[**MetricGenerators**<br>- BNC Generator<br>- CNSB Generator<br>- Latency Generator]
            C2["**MetricCombiner**<br>(WeightedCombiner)"]
            C3[**ScoreTransformer**<br>- EWMA Transformer<br>- HighPass Transformer]
        end
    end

    classDef leftAlign text-align:left;
    class C1,C2,C3 leftAlign;
```
(Don't get me wrong - I do like ascii art as it's viewable in the terminal, but for the other diagrams, not so much).

smoyer64 (approved):
Amazing work ... I have minor (due to their use) concerns about not checking for the integer overflows but I can ignore those if the team agrees.

smoyer64:
Do we want any observability in the dynamic load balancer?  What metrics would be useful?

lib/watcherscore/consts.go
noleto: Yes, in a future version, we plan to expose those parameters directly in the Caddyfile (or DIN Registry). For now, we’ve kept it simple.

lib/watcherscore/transformers_test.go
noleto: good catch, fixed!

lib/watcherscore/transformation_test.go
noleto: improved

lib/watcherscore/types.go
noleto: agree, immutable structures make it easier to design concurrent programs, so in general, it’s my default choice. (By the way, this is one reason I prefer Rust, which is immutable by default.)

lib/watcherscore/watcherscore_manager.go
noleto: Yes, this is the intended behavior (networks operate independently of one another, and an error in one should not stop the process). I’ve added error handling to log the error explicitly, good catch indeed!

modules/caddy_unmarshaller.go
noleto: Yeah, this is an important point. I replaced it with `strconv.ParseUint(...)` to parse unsigned long integers only. Safer code, less headaches!

docs/dynamic-load-balancing/dynamic-load-balancing.md
noleto: Indeed, I was experimenting with Mermaid generation after seeing some comments in the Slack channel. This one was generated by Claude, so it’s in ASCII.

noleto:
> Do we want any observability in the dynamic load balancer? What metrics would be useful?

Yes, this why you can see that all logs have two prefixes:
* [WATCHER_SCORE]... for all loggers related to watcher score logic
* [DYNAMIC_LB]... for all loggers related to the DIN proxy logic that uses provider scores to dynamic load balance the traffic
So I can capture metrics in SigNoz

smoyer64:
> > Do we want any observability in the dynamic load balancer? What metrics would be useful?
> 
> Yes, this why you can see that all logs have two prefixes:
> 
>     * [WATCHER_SCORE]... for all loggers related to watcher score logic
> 
>     * [DYNAMIC_LB]... for all loggers related to the DIN proxy logic that uses provider scores to dynamic load balance the traffic
>       So I can capture metrics in SigNoz

Yeah, I saw all the logging but I was wondering about the other two signals (metrics and traces.)

modules/caddy_unmarshaller.go
smoyer64: Since you may have external configuration values for these later, you could instead call it `WatcherScoreSyncInterval` and accept the constant or configuration value as a `string`.  Then use `time.ParseDuration` which would leave you with nice looking configuration values (like `5m`, 10s`, etc)

PR #153, opened 2025-10-31
Chore/align builds, support podman

smoyer64:
This simple PR performs the following maintenance on the `din-caddy-plugins` repository:

1. Updates `go.mod` (and `go.sum`) to use the same versions of the `caddy` and `xcaddy` packages that are being used in the Dockerfile.
2. Updates `Dockerfile` and `compose.yaml` (and as a side effect, the `Makefile` to use fully qualified image names (including the registry).  This theoretically allows any OCI-compliant container environment to build and run our custom Caddy server.
3. Fixed the location of of the global `log.format` directive in `Caddyfile` - it was breaking the expected behavior of `make run`.

Regarding bullet #3, I've tested the change both running locally and via `docker-compose` (for me using `podman-compose` compatibility mode).  The logs are now emitted as expected for both execution environments, but it nags me that the the Docker image seems to run with the `log.format` directive in the wrong place (but the local server fails to start).


Makefile
```diff
@@ -206,31 +206,6 @@ docs: ## Generate documentation
 	@echo "$(GREEN)Generating documentation...$(NC)"
 	go doc -all ./... > docs/API.md
 
-## Secret Management Commands
```
natefikru, line 209: is there a reason you are taking these out?

natefikru:
everything looks good, just have the question about roming the auto caddyfile template secrete generator command

natefikru:
everything looks good, just have the question about removing the auto caddyfile template secret generator commands.

Makefile
smoyer64: No - I've replace them with the latest commit

natefikru (approved):

PR #156, opened 2025-11-06
Update archive checks to use eth_getBalance instead of eth_call

AusIV:
Some EVM-like networks (notably zk-sync) behave a bit differently for eth_call on older block ranges. After some discussion, we decided that we didn't want to make a special exception for this network, and that eth_getBalance is a more universal way to check for state availability on archive blocks.


noleto (approved):
LGTM

PR #157, opened 2025-11-06
Reduce Prometheus Histogram Metrics Collection via Sampling

natefikru:
### Problem
High Prometheus metrics bill due to expensive histogram metric collection with high cardinality. Previous approach made all-or-nothing decisions per label group - if `["bitcoin-mainnet", "liquify", "eth_call", "200"]` hashed above the 25% threshold, ALL requests with those labels were excluded, potentially missing entire endpoints.

#### Before (All-or-Nothing)
```
Labels: ["bitcoin-mainnet", "liquify", "eth_call", "200"]
Request 1: ❌ Not sampled (hash above threshold)
Request 2: ❌ Not sampled (same hash)
Request 3: ❌ Not sampled (same hash)
Request 4: ❌ Not sampled (same hash)
Result: 0% sampling - missing entire endpoint!
```

#### After (Within-Group Sampling)
```
Labels: ["bitcoin-mainnet", "liquify", "eth_call", "200"]
Request 1: ✅ Sampled (hash with timestamp1 below threshold)
Request 2: ❌ Not sampled (hash with timestamp2 above threshold)
Request 3: ❌ Not sampled (hash with timestamp3 above threshold)
Request 4: ✅ Sampled (hash with timestamp4 below threshold)
Result: 25% sampling - proper coverage!
```

### Solution
Implemented **timestamp-based hash sampling** that ensures proper sampling WITHIN each label group:

#### Sampling Strategy
- **Per-Request Uniqueness**: Hash combines labels + nanosecond timestamp
- **Within-Group Sampling**: Each request gets independent sampling decision (25% or 50% based on metric type)
- **Error Boosting**: 100% sampling for errors (HTTP 4xx/5xx, Warning/Unhealthy status)
- **Simple Implementation**: Just added timestamp to existing hash function
- **Stateless**: No memory or state management needed

#### Sampling Rates
- **`DinRequestDurationMilliseconds`**: 25% normal requests, 100% errors
- **`DinNetworkRequestHealthCheckDurationMilliseconds`**: 50% normal health checks, 100% errors

#### New Files
- `lib/prometheus/sampler.go`: HybridSampler with timestamp-based hash sampling for within-group sampling
- `lib/prometheus/sampler_test.go`: Tests for within-group sampling rates and error boosting

#### Modified Files
- `lib/prometheus/prometheus.go`:
  - Added `requestSampler` and `healthCheckSampler` to PrometheusClient
  - Replaced `rand.Intn()` with timestamp-based sampling
  - Uses `ShouldSampleRequest()` and `ShouldSampleHealthCheck()` for intelligent sampling
  - Properly handles HealthStatus enum values ("Healthy", "Warning", "Unhealthy")
  - Simplified timestamp handling using `fmt.Sprintf("%d", timestamp)` for better code clarity
  - Replaced complex byte operations with straightforward string formatting

### Technical Benefits
- **True Within-Group Sampling**: Each request gets independent decision, not all-or-nothing per group
- **Complete Coverage**: All label combinations get sampled over time
- **Error Capture**: 100% sampling of errors ensures no failures are missed
- **Simple & Fast**: Just added timestamp to hash - minimal code change, low overhead
- **Statistically Valid**: Achieves accurate 25% or 50% sampling rates within each group

### DIFF
--- a/lib/prometheus/prometheus.go
+++ b/lib/prometheus/prometheus.go
@@ -1,6 +1,7 @@
 package prometheus
 
 import (
+	"math/rand"
 	"strconv"
 	"strings"
 	"time"
@@ -156,8 +157,10 @@ func (p *PrometheusClient) HandleRequestMetrics(data *PromRequestMetricData, dur
 	// Increment prometheus counter metric based on request data
 	DinRequestCount.WithLabelValues(network, method, data.Provider, data.ProviderName, data.ApiKey, data.HostName, status, data.HealthStatus, p.machineID, data.Environment).Inc()
 
-	// Observe prometheus histogram based on request duration and data
-	DinRequestDurationMilliseconds.WithLabelValues(network, method, data.Provider, data.ProviderName, data.HostName, status, data.HealthStatus, p.machineID, data.Environment).Observe(float64(durationMS))
+	// Sample 25% of requests for histogram metrics to reduce costs (histograms are expensive)
+	if rand.Intn(4) == 0 {
+		DinRequestDurationMilliseconds.WithLabelValues(network, method, data.Provider, data.ProviderName, data.HostName, status, data.HealthStatus, p.machineID, data.Environment).Observe(float64(durationMS))
+	}
 }
 
 type PromHealthCheckMetricData struct {
@@ -207,5 +210,9 @@ func (p *PrometheusClient) HandleNetworkHealthCheckMetric(data *PromNetworkHealt
 	)
 
 	DinNetworkHealthCheckCount.WithLabelValues(network, status, p.machineID, data.Environment).Inc()
-	DinNetworkRequestHealthCheckDurationMilliseconds.WithLabelValues(network, status, p.machineID, data.Environment).Observe(float64(durationMS))
+
+	// Sample 50% of health checks for histogram metrics to reduce costs
+	if rand.Intn(2) == 0 {
+		DinNetworkRequestHealthCheckDurationMilliseconds.WithLabelValues(network, status, p.machineID, data.Environment).Observe(float64(durationMS))
+	}
 }
--- a/lib/prometheus/prometheus.go
+++ b/lib/prometheus/prometheus.go
@@ -1,7 +1,6 @@
 package prometheus
 
 import (
-	"math/rand"
 	"strconv"
 	"strings"
 	"time"
@@ -26,15 +25,19 @@ const (
 
 // PrometheusClient is a struct that holds the prometheus client
 type PrometheusClient struct {
-	logger    *logger.LoggerClient
-	machineID string
+	logger             *logger.LoggerClient
+	machineID          string
+	requestSampler     *HybridSampler
+	healthCheckSampler *HybridSampler
 }
 
 // NewPrometheusClient returns a new prometheus client
 func NewPrometheusClient(logger *logger.LoggerClient, machineId string) *PrometheusClient {
 	return &PrometheusClient{
-		logger:    logger,
-		machineID: machineId,
+		logger:             logger,
+		machineID:          machineId,
+		requestSampler:     NewHybridSampler(0.25, 1.0), // 25% for normal requests, 100% for errors
+		healthCheckSampler: NewHybridSampler(0.5, 1.0),  // 50% for normal health checks, 100% for errors
 	}
 }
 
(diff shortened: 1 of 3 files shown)


lib/prometheus/prometheus.go
```diff
@@ -207,5 +210,9 @@ func (p *PrometheusClient) HandleNetworkHealthCheckMetric(data *PromNetworkHealt
 	)
 
 	DinNetworkHealthCheckCount.WithLabelValues(network, status, p.machineID, data.Environment).Inc()
-	DinNetworkRequestHealthCheckDurationMilliseconds.WithLabelValues(network, status, p.machineID, data.Environment).Observe(float64(durationMS))
+
+	// Sample 50% of health checks for histogram metrics to reduce costs
+	if rand.Intn(2) == 0 {
+		DinNetworkRequestHealthCheckDurationMilliseconds.WithLabelValues(network, status, p.machineID, data.Environment).Observe(float64(durationMS))
```
noleto: I understand that this is a quick win. A more robust solution would be to implement a single process for collecting health check metrics, external to the DIN Router (which is deployed across multiple instances).

noleto: We may need to review the thresholds currently used in alerting and monitoring (as of now, we use the count of health checks)

lib/prometheus/prometheus.go
```diff
@@ -156,8 +157,10 @@ func (p *PrometheusClient) HandleRequestMetrics(data *PromRequestMetricData, dur
 	// Increment prometheus counter metric based on request data
 	DinRequestCount.WithLabelValues(network, method, data.Provider, data.ProviderName, data.ApiKey, data.HostName, status, data.HealthStatus, p.machineID, data.Environment).Inc()
 
-	// Observe prometheus histogram based on request duration and data
-	DinRequestDurationMilliseconds.WithLabelValues(network, method, data.Provider, data.ProviderName, data.HostName, status, data.HealthStatus, p.machineID, data.Environment).Observe(float64(durationMS))
+	// Sample 25% of requests for histogram metrics to reduce costs (histograms are expensive)
+	if rand.Intn(4) == 0 {
```
noleto: Ideally, this sampling should be done by network/method to ensure the proper distribution of methods when calculating latency per method (using stratified sampling). However, I understand that this is a quick and urgent task for billing purposes, but please keep in mind that it may distort analytics based on network/provider/method.

noleto:
> * **Statistical Validity**: Percentiles (p50, p95, p99) remain accurate with sampling

Not entirely true: percentiles remain accurate when used without dimensions (i.e., without breaking down by method, provider, or network). However, as soon as you break it down by any dimension, the lack of stratification in sampling will distort the percentiles!

smoyer64:
OT: I was expecting this to be `opentelemetry-collector-contrib` but I haven't done Prometheus in a long while and shouldn't have an opinion.  I am curious about why we're using Prometheus for metrics instead of OTEL though.

noleto:
> > * **Statistical Validity**: Percentiles (p50, p95, p99) remain accurate with sampling
 
Here is an impact analysis created with help of ChatGPT 5:
image

lib/prometheus/prometheus.go
```diff
@@ -156,8 +157,10 @@ func (p *PrometheusClient) HandleRequestMetrics(data *PromRequestMetricData, dur
...
+		DinRequestDurationMilliseconds.WithLabelValues(network, method, data.Provider, data.ProviderName, data.HostName, status, data.HealthStatus, p.machineID, data.Environment).Observe(float64(durationMS))
```
noleto, line 166: I believe we could just remove these labels: `providerName`, `hostName`, `machineID`

natefikru:
> OT: I was expecting this to be `opentelemetry-collector-contrib` but I haven't done Prometheus in a long while and shouldn't have an opinion. I am curious about why we're using Prometheus for metrics instead of OTEL though.

we are actually feeding these prometheus metrics directly to an open telemetry collector that feeds it to signoz. this is being done for logs as well.

natefikru:
> > > * **Statistical Validity**: Percentiles (p50, p95, p99) remain accurate with sampling
> 
> Here is an impact analysis created with help of ChatGPT 5: image

we do counts via a different metric.

lib/prometheus/prometheus.go
natefikru: updated via sampler function

lib/prometheus/prometheus.go
natefikru: updated via sampler function

lib/prometheus/sampler.go
```diff
@@ -0,0 +1,81 @@
...
+	errorRate float64 // Higher sampling rate for errors
+}
+
+// NewHybridSampler creates a new hybrid sampler with specified rates.
+// baseRate: sampling rate for normal requests (e.g., 0.25 for 25%)
+// errorRate: sampling rate for errors (e.g., 1.0 for 100%)
+func NewHybridSampler(baseRate, errorRate float64) *HybridSampler {
+	return &HybridSampler{
+		baseRate:  baseRate,
+		errorRate: errorRate,
```
noleto, line 38: We should enforce rate boundaries within [0.0, 1.0], as negative values or rates exceeding 1.0 have no valid meaning (the latter could indicate oversampling).

lib/prometheus/sampler.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// Always sample if rate is 1.0 or higher
+	if rate >= 1.0 {
+		return true
+	}
+
+	// Never sample if rate is 0 or negative
+	if rate <= 0 {
+		return false
+	}
```
noleto: If we enforce boundaries [0.0, 1.0], these checks will no longer be necessary. Additionally, it’s unclear, does rate >= 1.0 indicate oversampling?

lib/prometheus/sampler.go
```diff
@@ -0,0 +1,81 @@
...
+		h.Write([]byte(label))
+		// Add separator to prevent collisions between adjacent labels
+		h.Write([]byte("|"))
+	}
+
+	hashValue := h.Sum64()
+
+	// Convert rate to threshold value
+	// We use the full uint64 range for maximum precision
+	threshold := uint64(float64(^uint64(0)) * rate)
```
noleto: I think this is more readable:  `math.MaxUint64` (instead of `^uint64(0)`)

noleto (commented):
I didn’t fully understand why we are hashing based on label combinations. Could you elaborate? I might have missed something, but using deterministic hashes could result in some combinations never being sampled (except for errors).

noleto (approved):
LGTM

Issue #164, opened 2025-12-03
Middleware Locks Causing Performance Issues


AusIV:
The DIN Middleware includes an RWMutex to protect concurrent access of certain resources, though what specific resources are protected by it are not clearly enumerated. I believe the primary resources protected by the lock are the `middleware.Networks` map and the provider maps embedded in it.

At present, the RLock is acquired by every request handled by the server, at the top of the ServeHTTP() method. It is acquired at the top of the method, and released by a deferred RUnlock(). As a result, the RLock is held for the life of each request.

This means that when a goroutine acquires the write lock, any new requests are blocked until:

* All outstanding requests release their RLocks
* The goroutine acquiring the write lock completes its task and releases the lock

Even if the goroutine acquiring the write lock is very fast, outstanding requests holding the lock may take up to several seconds to complete, in which case the blocked requests will not even begin until all outstanding requests are complete.

---

I believe we need to take a two pronged approach here: 

## Releasing RLocks During Request Execution

I need to do a deeper analysis, but at first glance it looks to me like releasing the RLock before calling the middleware's `next()` function and re-acquiring it immediately afterwards would avoid interactions with the resources that need to be protected by a lock, while ensuring that Lock acquisitions by write operations are not dependent on provider response times or the complexity of requests.

### Registry Sync

When @natefikru originally implemented the DIN Registry integration, I remember reviewing the critical sections of that code and making sure they were about as minimal as they could be. If there are any gains to be made for the DIN Registry Sync, it will be in updates made after that initial push.

### Watcher sync

This issue was turned up when we enabled the Watcher Sync and had a performance impact as a result. Since that time, I've only taken a cursory look at the code sections protected by the lock, but I believe it is making multiple network calls to the Watcher API while the lock is held. We should try to minimize this to make the watcher calls outside of the critical section, and only update the middleware's datastructures during the critical sections.

That said, I'm also not sure the Watcher Sync is actually altering any of the data structures that need to be protected by the lock. I think we might be able to get away with the Watcher Sync acquiring a read lock instead of a write lock. I recognize that it changes data, but my first pass makes me think that the data it's updating is not the data we need the locks to protect.

noleto:
> ## Releasing RLocks During Request Execution
> ...it looks to me like releasing the RLock before calling the middleware's `next()` function and re-acquiring it immediately afterwards would avoid interactions with the resources that need to be protected by a lock

[...]

PR #168, opened 2025-12-09
Fine-Granular lock to only shared fields 

noleto:
This pull request addresses issue #164 by narrowing the scope of locks to only include shared resources. Previously, we had a global lock at the middleware level, but for Watcher syncing scores, we only update a single score at the provider level.

This code takes a very granular approach by protecting only the score field being updated. Please note that this fix does not address the issues with the DIN Registry code, which is more complex and involves changes to the `network` map (DIN Registry may add or remove networks or providers). A separate pull request will be created to address those broader concerns with a comprehensive solution. This fix specifically addresses the Watcher sync score and its interaction with the global lock in `ServeHTTP` that was previously  being held for too long. 

EDIT:
I've also revamped the computation logic for the watcher manager to utilize the same approach with fine-grained locks. Additionally, I've ensured that computations don't stack up if the `ComputeScores()` method is called multiple times by concurrent threads.  I added tests to verify that the concurrency in score computation works as intended.

### DIFF
--- a/lib/watcherscore/interface.go
+++ b/lib/watcherscore/interface.go
@@ -9,19 +9,27 @@ import (
 // IWatcherScoreManager defines the responsibilities of the component that computes and stores the scores for providers on a given network.
 type IWatcherScoreManager interface {
 	// ComputeScores computes the scores for all providers on all networks
+	// It is safe to call this method concurrently
+	// Note that once a computation is in progress, additional calls will be skipped and the method will return immediately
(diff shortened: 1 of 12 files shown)


smoyer64 (approved):

natefikru:
lgtm

natefikru (approved):

smoyer64 (approved):

natefikru (approved):

PR #169, opened 2025-12-11
Add debug_traceBlockByNumber archive check for MetaMask compliance

natefikru:
## Summary

Adds an optional `debug_traceBlockByNumber` check to the archive health check flow for MetaMask compliance requirements. This verifies that providers support trace/debug capabilities in addition to the existing `eth_getBalance` archive check.

### New Configuration

```caddyfile
networks {
    eth {
        archive_enabled true
        archive_trace_block_by_number true
    }
}
```

### Behavior

| archive_enabled | archive_trace_block_by_number | Result |
|-----------------|-------------------------------|--------|
| false | false | No archive checks |
| false | true | No archive checks |
| true | false | Only eth_getBalance |
| true | true | eth_getBalance + debug_traceBlockByNumber |

### Request Payload

```json
{
  "jsonrpc": "2.0",
  "method": "debug_traceBlockByNumber",
  "id": 1,
  "params": ["0xBLOCK_HEX", {"tracer": "callTracer", "timeout": "30s", "onlyTopCall": true}]
}
```

### Failure Handling

- Uses same block number as existing archive check (quarter-height: `currentBlock / 4`)
- Failures return **Warning** status (not Unhealthy), keeping provider in traffic rotation
- Properly detects JSON-RPC errors (e.g., `-32000 block pruned`) even on HTTP 200 responses
- Failures logged via `logProviderWarning()` with `health_status=Warning` label

### Files Changed

- `modules/consts.go` - Added default constant
- `modules/network.go` - Added struct fields, flag tracking, trace check integration
- `modules/caddy_unmarshaller.go` - Added Caddyfile parsing
- `lib/network/json_rpc_client.go` - Added shared helper function
- `lib/network/evm_handler.go` - Added EVM-specific trace methods

### DIFF
--- a/lib/network/evm_handler.go
+++ b/lib/network/evm_handler.go
@@ -479,6 +479,42 @@ func (h *EVMHandler) PerformArchiveCheck(httpUrl string, headers map[string]stri
 	)
 }
 
+// PerformTraceBlockByNumberCheck performs debug_traceBlockByNumber check for EVM chains
+// This is an additional archive mode check that verifies trace capabilities (MetaMask requirement)
+func (h *EVMHandler) PerformTraceBlockByNumberCheck(httpUrl string, headers map[string]string, httpClient din_http.IHTTPClient, authClient auth.IAuthClient, requestAttempts int, blockHeight string) error {
+	return PerformTraceBlockByNumberCheckViaJSONRPC(
+		httpUrl,
+		headers,
+		httpClient,
+		authClient,
+		requestAttempts,
+		blockHeight,
+		h.CreateTraceBlockByNumberPayload,
+		h.ParseTraceBlockByNumberResponse,
+	)
+}
+
+// CreateTraceBlockByNumberPayload creates the debug_traceBlockByNumber request payload
+func (h *EVMHandler) CreateTraceBlockByNumberPayload(blockHeight string) ([]byte, error) {
+	payload := fmt.Sprintf(`{"jsonrpc":"2.0","method":"debug_traceBlockByNumber","id":1,"params":["%s",{"tracer":"callTracer","timeout":"30s","onlyTopCall":true}]}`, blockHeight)
+	return []byte(payload), nil
+}
+
+// ParseTraceBlockByNumberResponse validates the debug_traceBlockByNumber response
+func (h *EVMHandler) ParseTraceBlockByNumberResponse(body []byte) error {
+	var respObject map[string]interface{}
+	if err := json.Unmarshal(body, &respObject); err != nil {
+		return fmt.Errorf("failed to unmarshal trace response: %w", err)
+	}
+
+	// Check for JSON-RPC error
+	if errField, ok := respObject["error"]; ok && errField != nil {
+		return fmt.Errorf("debug_traceBlockByNumber not supported: %v", errField)
+	}
+
+	return nil
+}
+
 // PerformGetBlockByNumber performs get block by number operation for EVM chains using JSON-RPC
 func (h *EVMHandler) PerformGetBlockByNumber(httpUrl string, headers map[string]string, httpClient din_http.IHTTPClient, authClient auth.IAuthClient, requestAttempts int, blockNumber int64) (interface{}, error) {
 	// Use the shared JSON-RPC helper for get block by number operations
--- a/lib/network/evm_handler.go
+++ b/lib/network/evm_handler.go
@@ -489,19 +489,17 @@ func (h *EVMHandler) PerformTraceBlockByNumberCheck(httpUrl string, headers map[
 		authClient,
 		requestAttempts,
 		blockHeight,
-		h.CreateTraceBlockByNumberPayload,
-		h.ParseTraceBlockByNumberResponse,
 	)
 }
 
-// CreateTraceBlockByNumberPayload creates the debug_traceBlockByNumber request payload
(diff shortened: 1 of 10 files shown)


lib/network/evm_handler.go
```diff
@@ -468,6 +468,42 @@ func (h *EVMHandler) PerformArchiveCheck(httpUrl string, headers map[string]stri
 	)
 }
 
+// PerformTraceBlockByNumberCheck performs debug_traceBlockByNumber check for EVM chains
+// This is an additional archive mode check that verifies trace capabilities (MetaMask requirement)
+func (h *EVMHandler) PerformTraceBlockByNumberCheck(httpUrl string, headers map[string]string, httpClient din_http.IHTTPClient, authClient auth.IAuthClient, requestAttempts int, blockHeight string) error {
```
noleto, line 484: I think it makes sense to add this method to the `NetworkHandler interface` contract so that every handler can decide whether it is eligible to implement it or not

lib/network/evm_handler.go
```diff
@@ -468,6 +468,42 @@ func (h *EVMHandler) PerformArchiveCheck(httpUrl string, headers map[string]stri
...
+	}
+
+	// Check for JSON-RPC error
+	if errField, ok := respObject["error"]; ok && errField != nil {
+		return fmt.Errorf("debug_traceBlockByNumber not supported: %v", errField)
+	}
+
+	return nil
+}
+
```
noleto: These two methods seem to be just helper functions, so we don't need to add them as handler methods. This will help keep the EVM handler more focused on implementing only the methods in the contract. Additionally, it would be great to move them to a helper file.

lib/network/json_rpc_client.go
```diff
@@ -179,6 +179,55 @@ func PerformArchiveCheckViaJSONRPC(httpUrl string, headers map[string]string, ht
 	return fmt.Errorf("failed after %d attempts: %w", requestAttempts, lastErr)
 }
 
+// PerformTraceBlockByNumberCheckViaJSONRPC performs a debug_traceBlockByNumber check via JSON-RPC
+// This is EVM-specific and verifies trace/debug capabilities for MetaMask compliance
+func PerformTraceBlockByNumberCheckViaJSONRPC(httpUrl string, headers map[string]string, httpClient din_http.IHTTPClient, authClient auth.IAuthClient, requestAttempts int, blockHeight string, createPayloadFunc func(string) ([]byte, error), parseResponseFunc func([]byte) error) error {
```
noleto: What is the rationale behind passing the closures `createPayloadFunc` and `parseResponseFunc` to this function? Does it vary depending on the caller? Currently, there is only a single call made passing `CreateTraceBlockByNumberPayload` and `ParseTraceBlockByNumberResponse`, respectively. If the behavior is likely to change depending on the caller, it might be better to introduce an interface to capture both of these behaviors you want to customize

noleto: see https://refactoring.guru/design-patterns/template-method/go/example

modules/network.go
```diff
@@ -444,7 +447,26 @@ func (n *network) performArchiveCheck(provider *provider, currentBlock int64) er
...
+		return err
+	}
+
+	// Additional trace check (EVM-specific, MetaMask requirement)
+	// Only runs if archive_trace_block_by_number flag is enabled
+	if n.ArchiveTraceBlockByNumberEnabled {
+		// Check if handler supports trace check via type assertion (EVM-specific)
+		type traceChecker interface {
+			PerformTraceBlockByNumberCheck(httpUrl string, headers map[string]string, httpClient din_http.IHTTPClient, authClient auth.IAuthClient, requestAttempts int, blockHeight string) error
+		}
```
noleto: I'd suggest moving the interface declaration to `handlers.go` as an optional interface to implement. Alternatively, and this is my preference, include it in the `NetworkHandler`, which will make the method discoverable from the `NetworkHandler` interface and eliminate the need for type assertions, resulting in more idiomatic Go code. In both cases, it will be easier to mock and test that behavior.

lib/network/evm_handler.go
natefikru: I would push against this because this is more of an evm specific call, it is a sub function of the archive check function which has an interface associated with it.

lib/network/json_rpc_client.go
natefikru: you're right, thats definitely over engineered, done

lib/network/evm_handler.go
natefikru: good call, done.

lib/network/evm_handler.go
smoyer64: I think we should abstract the idea of healthcheck even more - I'll have a proposal that shrinks the `Handler` interface in January

smoyer64 (approved):

lib/network/evm_handler.go
noleto: Okay, it's fair if that's specific to EVM. It should at least have its own interface defined in the `handlers.go` file, as you currently have it declared inline here: https://github.com/DIN-center/din-caddy-plugins/blob/0aab16f490f91699c5d88fd2dce29d8e0ffe5883/modules/network.go#L459-L461) This is not to be picky: it makes the implementation discoverable by IDE tools and also makes it easier to mock the behavior in tests. Defining an inline interface, while possible, is not very idiomatic in Go code

noleto (approved):
looks good to me, thanks for addressing my feedback.

PR #171, opened 2025-12-12
Fix provider URL query parameter handling

natefikru:
## Summary

Provider endpoint URLs with query parameters (e.g., `https://lb-penguin.nodies.app/v2/sei?apikey=xxx`) were not properly supported. Query parameters were being stripped during request proxying, though health checks worked correctly.

### Root Cause
- `din_middleware.go`: Only stored `url.Path`, discarding `url.RawQuery`
- `din_select.go`: Only set request path when forwarding, didn't update query params

### Changes
- Added `query` field to provider struct to store URL query string
- Extract `url.RawQuery` during provider initialization in `initializeProvider`
- Merge provider query params with request query params when proxying in `Select`
- Provider query params take precedence (placed first in merged query string)
- Updated debug logging to include query params

### Test Coverage
- Added tests for URLs with query params (apikey in query string)
- Added tests for URLs with multiple query params
- Added tests for URLs with API keys in both path and query
- Added tests for query param merging logic
- Backwards compatibility tests for URLs without query params

## Test Plan

- [x] Unit tests pass (`go test ./...`)
- [x] Test provider URL with query params is parsed correctly
- [x] Test query params are preserved when proxying
- [x] Test provider query params merge correctly with incoming request query params
- [x] Test empty query params don't add spurious `?` or `&`
- [x] Backwards compatibility: URLs without query params still work

### DIFF
--- a/modules/din_select.go
+++ b/modules/din_select.go
@@ -102,6 +102,16 @@ func (d *DinSelect) applyProviderConfiguration(provider *provider, r *http.Reque
 		}
 	}
 
+	// Merge provider query params with request query params
+	// Provider query params take precedence (placed first)
+	if provider.query != "" {
+		if r.URL.RawQuery == "" {
+			r.URL.RawQuery = provider.query
+		} else {
+			r.URL.RawQuery = provider.query + "&" + r.URL.RawQuery
+		}
+	}
+
 	// Apply headers
 	for k, v := range provider.Headers {
 		r.Header.Add(k, v)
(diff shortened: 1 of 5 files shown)


modules/din_select.go
```diff
@@ -102,6 +102,16 @@ func (d *DinSelect) applyProviderConfiguration(provider *provider, r *http.Reque
 		}
 	}
 
+	// Merge provider query params with request query params
```
AusIV: This seems like it should happen within the `ConfigureRequestPath` handler, as it could definitely have different ramifications for REST-based endpoints than it does for RPC based endpoints. I think right now if we just put it in the EVM handler we can cover today's scenario, and we can evaluate the ramifications for other networks when we're not rushed.

AusIV (approved):

PR #175, opened 2025-12-17
feat(docker): fix log ingest in LGTM stack

smoyer64:
While developing the metric creation for the "Anomaly detection based error rate" requirement in "SignOz update based on Issue 1390", I noticed that log ingestion wasn't working when running Caddy with the LGTM stack.  This happens because the Caddy server, running as root, creates the `caddy.log` file with very restrictive permissions.

The solution is to run both the `din-caddy` image and the `opentelemetry-collector-contrib` image using the host OS user's `UID` and `GID`.  The syntax for running this is a little awkward to continuously type, so I also added the `docker-compose` `Make` recipe.

Now run Caddy with the LGTM stack using the following command:

```
make docker-compose
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 # Ensure bash is used for all shell commands to guarantee cross-environment reproducibility
 SHELL := /bin/bash
 
-.PHONY: help build run test test-verbose test-coverage test-race clean secure dev-deps lint format check-deps benchmark profile docker-build docker-run
+.PHONY: help build run test test-verbose test-coverage test-race clean secure dev-deps lint format check-deps benchmark profile docker-build docker-run docker-compose
 
 # Default target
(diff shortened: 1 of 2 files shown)


noleto:
Just to double check, this fixes only for running locally right? because when deployed in prod, we don't use this `compose.yml` file (we generates a new one => https://github.com/DIN-center/din-gateway-deployment/blob/main/startup-script.sh#L110)

smoyer64:
> Just to double check, this fixes only for running locally right? because when deployed in prod, we don't use this `compose.yml` file (we generates a new one => https://github.com/DIN-center/din-gateway-deployment/blob/main/startup-script.sh#L110)

Yes, this is only for running locally

noleto (approved):
LGTM

PR #176, opened 2025-12-17
Dynamic Block Lag Limit Calculation

natefikru:
## Summary

Adds automatic calculation of block lag limits based on actual blockchain block production rates. Different blockchain networks produce blocks at varying speeds (e.g., Solana ~400ms, BSC ~3s, Starknet ~2s, Bitcoin ~10min), so a static block lag limit is suboptimal for provider health monitoring. This feature calculates the appropriate limit for each network on startup.

## How It Works

On network startup, the system:
1. Retrieves the latest block number from a healthy provider
2. Fetches timestamps for two blocks: the latest block and a block 1024 blocks earlier
3. Calculates average block time from the timestamp difference
4. Computes the optimal block lag limit: `13000ms / avg_block_time_ms`
5. Rounds up to the nearest multiple of 5, with a minimum of 5 blocks

The calculation uses immutable blockchain timestamps, ensuring deterministic results across all server instances querying the same blocks. Using 1024 blocks for the lookback provides accurate measurement and makes reorgs/delays a rounding error.

### Automatic - No Configuration Required

This feature is **completely automatic** and requires no Caddyfile changes:

1. ✅ Enabled by default for all networks
2. ✅ Runs once at network startup (asynchronously)
3. ✅ Automatically detects if handler supports dynamic calculation
4. ✅ Falls back to default (15 blocks) if calculation fails

### What You'll See in Logs

**Successful calculation** (INFO level):
```
Dynamic block lag limit calculated and applied
  network: ethereum/sepolia
  old_limit: 15
  new_limit: 5
  avg_block_time_ms: 2400.5
  lookback_blocks: 1024
```

**Handler doesn't support** (DEBUG level):
```
Handler does not support dynamic block lag, keeping default
  network: beacon/mainnet
  default_limit: 15
```

**Calculation failed** (WARN level):
```
Failed to measure block time, keeping default
  network: bsc/mainnet
  default_limit: 15
  error: [details]
```

### Verifying It's Working

Check your Caddy logs after startup:
- Supported networks (EVM, Solana, Starknet, Bitcoin) will show **"Dynamic block lag limit calculated"**
- Unsupported networks (Beacon Chain) will show **"keeping default"**

### Optional: Override with Caddyfile

You can still manually configure block lag limit if needed:

```caddyfile
din {
    networks {
        ethereum/sepolia {
            healthcheck_blocklag_limit 10  # Override dynamic calculation
            # ... rest of config
        }
    }
}
```

**Note**: Manual configuration takes precedence and disables dynamic calculation for that network.

### Supported Networks

Networks with dynamic calculation support:

| Network Type | Typical Block Time | Calculated Limit (13s period) |
|--------------|-------------------|-------------------------------|
| Ethereum     | ~12s              | 5 blocks (minimum)            |
| BSC          | ~3s               | 5 blocks                      |
| Polygon      | ~2s               | 10 blocks                     |
| Arbitrum     | ~0.25s            | 55 blocks                     |
| Optimism     | ~2s               | 10 blocks                     |
| Solana       | ~0.4s             | 35 blocks                     |
| Starknet     | ~2s               | 10 blocks                     |
| Bitcoin      | ~10min            | 5 blocks (minimum)            |

Networks using default (15 blocks):
- Beacon Chain (slots/epochs model)
- Bitcoin Esplora (REST API limitations)
- Any network where calculation fails

## Supported Handlers

- **EVM** (Ethereum, BSC, zkSync, SEI, etc.)
- **Solana**
- **Starknet**
- **Bitcoin**

Handlers that don't support dynamic calculation (Beacon Chain, Bitcoin Esplora) retain their configured default limits.

## Interface Additions

Two methods added to `NetworkHandler`:
- `SupportsDynamicBlockLag() bool` - indicates whether the handler supports dynamic calculation
- `GetBlockTimestamp(blockNumber int64) (int64, error)` - retrieves Unix timestamp for a specific block

## Default Block Lag Limit

The default block lag limit (`DefaultBlockLagLimit`) is set to **15 blocks**. This conservative default provides:
- Adequate tolerance for networks where dynamic calculation fails or is not supported
- Safety margin for EVM networks with variable block times during network congestion
- Coverage for archive networks that may have higher provider sync variance

For most EVM networks (~2s block time), dynamic calculation yields 5-10 blocks. For faster chains like Solana (~400ms), it yields ~30-35 blocks. The 15-block default serves as a safe middle ground when dynamic calculation is unavailable.

## Fallback Behavior

If the calculation fails (no providers available, network errors, etc.), the system retains the configured default `BlockLagLimit` value.

---

## Notes

**Immutable Timestamp-Based Calculation**: This implementation uses blockchain block timestamps rather than real-time observation to calculate block production rates. Block timestamps are immutable data baked into the blockchain itself, meaning any server querying blocks N and N-1024 will always get the same timestamps and therefore compute the same block lag limit. This ensures deterministic, consistent results across all server instances without requiring coordination or shared state.

---

*Supersedes PR #134*


lib/network/solana_handler.go
```diff
@@ -107,7 +105,7 @@ func (h *SolanaHandler) ValidateRequest(req *http.Request) error {
 
 	// Check method
 	if req.Method != "POST" {
-		return fmt.Errorf("the Solana JSON-RPC requires POST method, got %s", req.Method)
+		return fmt.Errorf("Solana JSON-RPC requires POST method, got %s", req.Method)
```
smoyer64, line 108: Was the word "the" there just because linters want error strings to start with a lowercase letter?

modules/network.go
```diff
@@ -170,6 +171,9 @@ func (n *network) UnmarshalJSON(data []byte) error {
 }
 
 func (n *network) startHealthcheck() {
+	// Start dynamic block lag limit calculation (runs once asynchronously)
+	go n.calculateDynamicBlockLagLimit()
```
smoyer64, line 175: If this is slow, will the healthchecks below use values that have not yet been set?

smoyer64 (approved):

modules/network.go
natefikru: Health checks just check the service directly they aren't subject to routing rules.

In the case of a normal request, the default value of 5 is used as the block lag placeholder until the initial block lag value has been determined.

lib/network/solana_handler.go
natefikru: Yeah just seems like an llm quirk

go.mod
AusIV, line 1: There are a lot of things being downgraded here. Was that intended?

AusIV:
Mostly looks good. Not sure the go.mod downgrades are intentional.

go.mod
natefikru: good call, that was because i was working from an old version of the code.

PR #178, opened 2026-01-15
Add API key to loopback calls

AusIV:
We're trying to narrow down the number of api-key-less calls coming in, and at this point most of them are probably loopback calls. Because of how data gets from the caddyfile to the loopback check, this ended up being a more invasive change than I expected, but I believe I got everything and some of the existing tests needed to be updated to expect the new header, so it's definitely coming through.


natefikru:
reviewed, it looks good to me.

natefikru (approved):

PR #179, opened 2026-01-16
feat: strip Authorization headers from incoming client requests

natefikru:
## Summary

Strip `Authorization` headers from incoming client requests to prevent interference with the proxy's authentication handling.

## Changes

- Strip Authorization header in `din_middleware.go` before forwarding to providers
- Add test coverage for header stripping

## Example

Requests like this will now have the Authorization header removed:
```bash
curl -X POST -H "Authorization: Bearer {INVALID_KEY}" \
    -d '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":1}' \
    https://proxy/optimism-mainnet
```

The header is stripped early in the middleware, following the same pattern as `Din-Api-Key` removal.

### DIFF
--- a/modules/din_middleware.go
+++ b/modules/din_middleware.go
@@ -561,6 +561,10 @@ func (d *DinMiddleware) ServeHTTP(rw http.ResponseWriter, r *http.Request, next 
 
 	// Strip Authorization header from client requests
 	// The proxy handles all authentication internally
+	// NOTE: This removes clients' ability to pass their own Authorization headers to providers.
+	// While there is no current use case for this capability, if future requirements emerge
+	// for client-controlled authentication (e.g., provider-specific bearer tokens),
+	// this behavior will need to be re-evaluated.
 	if r.Header.Get("Authorization") != "" {
 		r.Header.Del("Authorization")
 	}
--- a/modules/din_middleware.go
+++ b/modules/din_middleware.go
@@ -565,9 +565,7 @@ func (d *DinMiddleware) ServeHTTP(rw http.ResponseWriter, r *http.Request, next 
 	// While there is no current use case for this capability, if future requirements emerge
 	// for client-controlled authentication (e.g., provider-specific bearer tokens),
 	// this behavior will need to be re-evaluated.
-	if r.Header.Get("Authorization") != "" {
-		r.Header.Del("Authorization")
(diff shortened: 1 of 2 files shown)


AusIV:
Why are we doing this? To my knowledge we don't use the Authorization header, nor do any of the protocols we support, so anyone sending it to us is doing so in error. If at some point in the future one of the protocols we support wants to use the Authorization header, this is going to make that difficult.

If we do have a good reason to want to strip the header, I'd advocate that we just do it with the Caddyfile, as discussed [in this thread](https://caddy.community/t/how-to-strip-headers/17638). The reason we don't strip the Din-Api-Key header via the Caddyfile is because we need to read it before it gets stripped.

AusIV (approved):
I still have some reservations, because when we add networks we often don't understand all the ways clients expect to be able to interact with networks. I see this coming up as "This feature of XYZ network isn't working, what is DIN doing here?"

Followed by several hours of troubleshooting before we even figure out the authorization header is involved.

Logging headers could help with this.

modules/din_middleware.go
```diff
@@ -559,6 +559,16 @@ func (d *DinMiddleware) ServeHTTP(rw http.ResponseWriter, r *http.Request, next
 		repl.Set("din_api_key", "unspecified")
 	}
 
+	// Strip Authorization header from client requests
+	// The proxy handles all authentication internally
+	// NOTE: This removes clients' ability to pass their own Authorization headers to providers.
+	// While there is no current use case for this capability, if future requirements emerge
+	// for client-controlled authentication (e.g., provider-specific bearer tokens),
+	// this behavior will need to be re-evaluated.
+	if r.Header.Get("Authorization") != "" {
```
smoyer64: This condition is not needed - just delete the header with that key without checking.

smoyer64 (commented):
@AusIV has a point but if we decide to clear this header in the plugin code, it can be reduced to just a single line.

natefikru:
@smoyer64 Good catch - removed the unnecessary condition in dffaec1. `Header.Del()` is safe to call unconditionally.

smoyer64:
> Header.Del() is safe to call unconditionally.

Yep - I even tested it to be sure ;)

PR #180, opened 2026-02-03
Fix CI startup delays with async health checks

natefikru:
## Summary

Makes health checks fully async with throttled concurrency (max 10 concurrent) to prevent blocking `Provision()` during startup.

## Problem

With 66+ networks, synchronous health checks during `Provision()` blocked for 160+ seconds, preventing the HTTP server on port 8000 from starting. This caused CI timeouts.

## Solution

Health checks now run in background goroutines with a semaphore limiting max 10 concurrent checks. `Provision()` completes in ~5 seconds while health checks run in the background.

### DIFF
--- a/modules/din_middleware.go
+++ b/modules/din_middleware.go
@@ -415,6 +415,7 @@ func (d *DinMiddleware) validateNetworkConfiguration(networkName string, network
 
 // startBackgroundServices starts health checks and registry sync
 func (d *DinMiddleware) startBackgroundServices() error {
+	// Start health checks
 	if err := d.startHealthChecks(); err != nil {
 		return fmt.Errorf("error starting healthchecks: %w", err)
 	}
--- a/modules/din_middleware.go
+++ b/modules/din_middleware.go
@@ -9,7 +9,6 @@ import (
 	"io"
 	"net/http"
 	"net/url"
-	"os"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


AusIV:
What is the behavior before the first healthcheck completes? Are things treated as healthy prior to the first completion, or unhealthy prior to the first completion?

If healthy, this could mean that the providers that are timing out still get routed to until the timeout fully fails (which could be a while).

If unhealthy, this could mean we're getting 503s before even healthy providers have had a chance to complete healthchecks. This is easier to deal with at the load balancer level though, as we don't start routing to a server until it has passed a couple of the load balancer health checks, so in practice healthy providers will have time to pass health checks before the instance is handling traffic.

Neither of these are ideal, but if default is unhealthy I think it's workable. I just want to make sure we understand the behavior.

natefikru:
> What is the behavior before the first healthcheck completes? Are things treated as healthy prior to the first completion, or unhealthy prior to the first completion?
> 
> If healthy, this could mean that the providers that are timing out still get routed to until the timeout fully fails (which could be a while).
> 
> If unhealthy, this could mean we're getting 503s before even healthy providers have had a chance to complete healthchecks. This is easier to deal with at the load balancer level though, as we don't start routing to a server until it has passed a couple of the load balancer health checks, so in practice healthy providers will have time to pass health checks before the instance is handling traffic.
> 
> Neither of these are ideal, but if default is unhealthy I think it's workable. I just want to make sure we understand the behavior.

Providers default to unhealthy before the first health check completes. `Healthy()` returns false when there's no block history yet

So yes, we'd get 503s if traffic hit the instance before health checks finish. But as you noted, the ALB health checks give us a buffer there. The instance won't receive traffic until it's passed a few probes, which should be enough time for the faster providers to complete their checks.

Agreed it's not ideal, but unhealthy by default feels like the safer choice between the two options.

AusIV (approved):

PR #181, opened 2026-02-09
feat: method-level failover for JSON-RPC -32601 errors

natefikru:
## Summary

- When a provider returns a JSON-RPC -32601 "method not found" error, the router now excludes that provider and retries on a different one
- Previously, -32601 was treated as non-retryable — the error was returned to the client immediately, even if another provider supported the method
- Matching uses the -32601 error code (provider-agnostic) with text pattern fallback
- Only applies to JSON-RPC handlers (EVM, Starknet, Solana, Bitcoin); REST handlers are unaffected
- If all providers return -32601, the original error is returned to the client

## Context

Raised by Iris — MegaETH's Alchemy provider didn't support `eth_getBlockReceipts` but QuickNode did. The router returned an error instead of failing over because `"method not found"` was classified as non-retryable.

Slack thread: https://consensys.slack.com/archives/C0A0XJB8NAH/p1770655485440739

## Known Limitations

- **HTTP 400 + -32601 body**: Some providers return HTTP 400 with a JSON-RPC -32601 body. The middleware classifies non-2xx responses as `"HTTP error: 400"` without parsing the body, so method-level failover won't trigger in this case. Most providers return HTTP 200 with the JSON-RPC error.
- **Batch JSON-RPC**: Batch responses with mixed errors are not individually classified (existing limitation).

### DIFF
--- a/lib/network/beacon_handler.go
+++ b/lib/network/beacon_handler.go
@@ -143,6 +143,11 @@ func (h *BeaconChainHandler) ParseResponse(body []byte, statusCode int) error {
 	return nil
 }
 
+func (h *BeaconChainHandler) IsRetryableOnDifferentProvider(err error, statusCode int) bool {
+	// REST APIs don't have JSON-RPC -32601 semantics
+	return false
+}
+
 func (h *BeaconChainHandler) IsRetryableError(err error, statusCode int) bool {
 	// HTTP server errors are retryable
 	if statusCode >= 500 {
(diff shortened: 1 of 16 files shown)


lib/network/beacon_handler.go
```diff
@@ -143,6 +143,11 @@ func (h *BeaconChainHandler) ParseResponse(body []byte, statusCode int) error {
 	return nil
 }
 
+func (h *BeaconChainHandler) IsRetryableOnDifferentProvider(err error, statusCode int) bool {
+	// REST APIs don't have JSON-RPC -32601 semantics
```
AusIV, line 147: Their equivalent is just a 404, but given the circumstances I think this is fine.

AusIV (approved):

Issue #184, opened 2026-02-12
Feature: Handler Required for the DIN Router for Stellar RPC Mainnet


twhay:
### Feature 1: Handler for Stellar RPC Mainnet

Stellar RPC is a lightweight tool that provides real-time access to Stellar network data. Much like RPC nodes in other blockchain ecosystems, it allows developers to query the network efficiently. - via [Stellar Documentation: RPC Introduction](https://developers.stellar.org/docs/data/apis/rpc)

The Stellar RPC API is a non-EVM chain, and therefore requires a specific handler. Below are the methods and a brief description of their functionality, for reference, as taken from the documentation:

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-324f8144-7fff-c851-133e-81dee12abdcd"><div dir="ltr" style="margin-left:0pt;" align="left">
Method Name | Description
-- | --
getEvents | Clients can request a filtered list of events emitted by a given ledger range.Stellar-RPC will support querying within a maximum of 7 days of recent ledgers.Note, this could be used by the client to only prompt a refresh when there is a new ledger with relevant events. It should also be used by backend Dapp components to "ingest" events into their own database for querying and serving.If making multiple requests, clients should deduplicate any events received, based on the event's unique id field. This prevents double-processing in the case of duplicate events being received.By default stellar-rpc retains the most recent 24 hours of events.getEvents lets you search for events that occurred over a ledger range. Events are emitted by the system and by smart contracts to communicate meaningful state changes to off chain indexers (see Events for more). Each event is made up of topics (which you can filter on) and data which are ScVals, Stellar's generic "value" type (see Contract Development for more).
getFeeStats | Statistics for charged inclusion fees. The inclusion fee statistics are calculated from the inclusion fees that were paid for the transactions to be included onto the ledger. For Soroban transactions and Stellar transactions, they each have their own inclusion fees and own surge pricing. Inclusion fees are used to prevent spam and prioritize transactions during network traffic surge.
getHealth | General node health check, will be primarily used to check the health of your nodes
getLatestLedger | For finding out the current latest known ledger of this node. This is a subset of the ledger info from Horizon.
[...]

PR #186, opened 2026-02-16
Add support for Stellar RPC

noleto:
### Add Stellar RPC Support

This PR tackles https://github.com/DIN-center/din-caddy-plugins/issues/184: It brings  Stellar RPC support to the DIN Caddy plugins, allowing the gateway to route and monitor Stellar RPC traffic. The implementation follows the existing handler pattern and includes full test coverage across the new Stellar-specific behavior.

### What’s Included

* **New Stellar RPC handler**
  Adds a dedicated handler at `lib/network/stellar_rpc_handler.go`.

* **Stellar JSON-RPC 2.0 compatibility**
  Implements Stellar’s JSON-RPC 2.0 request/response structure and method routing as defined in the Stellar RPC API spec => https://developers.stellar.org/docs/data/apis/rpc/api-reference/structure/json-rpc

* **Provider health checks via `getHealth`**
  Uses Stellar RPC’s native `getHealth` endpoint for liveness/readiness monitoring.

* **Ledger-first semantics (no “blocks”)**
  Stellar does not expose blocks in the same way as EVM chains. The handler maps:

  * “block operations” → **ledger operations** (via `getLedgers`)
  * “block number” → **latest ledger sequence** (via `getLatestLedger`)

* **Network validation using passphrases**
  Validates the **Stellar network passphrase** (Mainnet/Testnet) to prevent misrouting across networks.

* **Supported RPC methods**
  `getEvents`, `getFeeStats`, `getHealth`, `getLatestLedger`, `getLedgerEntries`, `getLedgers`, `getNetwork`, `getTransaction`, `getTransactions`, `getVersionInfo`, `sendTransaction`, `simulateTransaction`

### Architecture & Platform Improvements

* **Strongly-typed handlers**
  Handler kinds are now enforced via types (no more string-based identifiers).

* **Reusable “latest height” extraction**
  Refactors the “latest block/height” payload parsing so it can be reused across handlers (now compatible with Stellar’s latest ledger sequence).

* **Brotli decompression support (`content-encoding: br`)**
  Adds Brotli decoding for Stellar providers behind certain proxies/CDNs (e.g., Cloudflare) that return Brotli-compressed responses. Uses `github.com/andybalholm/brotli`.

* **Auto-generated mocks for `NetworkHandler`**
  Adds generated mocks to improve test ergonomics and keep interfaces consistent.

### How to test locally (Stellar RPC)

1. **Update `Caddyfile.private`** (public Stellar RPC providers only):

```caddyfile
stellar-mainnet-rpc {
  methods getEvents getFeeStats getHealth getLatestLedger getLedgerEntries getLedgers getNetwork getTransaction getTransactions getVersionInfo sendTransaction simulateTransaction
  providers {
    https://mainnet.sorobanrpc.com
    https://stellar-soroban-public.nodies.app
  }
  handler stellar-rpc
  chain_id "Public Global Stellar Network"
}

stellar-testnet-rpc {
  methods getEvents getFeeStats getHealth getLatestLedger getLedgerEntries getLedgers getNetwork getTransaction getTransactions getVersionInfo sendTransaction simulateTransaction
  providers {
    https://stellar.liquify.com/api=41EEWAH79Y5OCGI7/testnet
    https://soroban-rpc.testnet.stellar.gateway.fm
    https://stellar-soroban-testnet-public.nodies.app
  }
  handler stellar-rpc
  chain_id "Test SDF Network"
}
```

2. **Build + run the DIN Router**:

```bash
make docker-build && docker compose up -d
```

3. **Run Tom’s Stellar RPC smoke test script**: https://github.com/DIN-center/din-caddy-plugins/issues/184#issuecomment-3893908723

```bash
./stellar-rpc.sh http://127.0.0.1:8000/stellar-mainnet-rpc
```

4. **Expected output**

```text
Testing Stellar RPC endpoint: http://127.0.0.1:8000/stellar-mainnet-rpc

  OK   getHealth
  OK   getVersionInfo
  OK   getNetwork
  OK   getLatestLedger
  OK   getFeeStats
  OK   getEvents (method exists, response: error -32600)
  OK   getLedgerEntries
  OK   getLedgers (method exists, response: error -32600)
  OK   getTransactions (method exists, response: error -32600)
  OK   getTransaction
  OK   sendTransaction
  OK   simulateTransaction

--- Summary ---
Passed: 12
Failed: 0
All 12 Stellar RPC methods are supported.
```


lib/network/stellar_rpc_handler.go
```diff
@@ -0,0 +1,509 @@
...
+	}
+
+	if len(chainID) == 0 {
+		return fmt.Errorf("empty chain ID")
+	}
+
+	// Stellar uses network passphrase as chain ID
+	// Check for known network prefixes: "Public Global Stellar Network" or "Test SDF Network"
+	// This is more lenient and allows for variations in the full passphrase
+	validPrefixes := []string{
```
natefikru, line 142: This is peculiar and definitely different from all of the other networks, just want to double check that this is the best way to handle this, and that there isn't a single chain identifier

lib/network/testdata/stellar_get_latest_ledger_response.json
```diff
@@ -0,0 +1,11 @@
+{
```
natefikru, line 1: We can reduce count of files by having all text fixtures organized by handler

natefikru (approved):
Left some nitpick reviews but overall it looks solid to me, I think this PR probably could have been split into 2 tbh.

modules/helpers_test.go
cds-amal, line 1: @smoyer64 , if the suffix ends `_tests.go`, would it be leaking into production?

noleto: this was a renaming issue, I reverted the name to helpers_test.go

lib/network/stellar_rpc_handler.go
noleto: Each Stellar network has a unique identifier in the format **“[Network Name] ; [Month of Creation] [Year of Creation]”** also known as the *network passphrase*. This passphrase is used to verify that a transaction signed for one network isn’t submitted to another. More details here: [https://developers.stellar.org/docs/networks#network-passphrases](https://developers.stellar.org/docs/networks#network-passphrases)

Internally, Stellar also uses a **Network ID**, which is simply the SHA-256 hash of the network passphrase.

noleto: This raises a broader question about how our current model in the DIN Router works. It was originally designed with EVM networks in mind, then extended to EVM-compatible networks, and later expanded to other types of networks. As we continue adding services that may not even be blockchain networks, we’ll need to revisit how we handle metadata, in my opinion.

lib/network/testdata/stellar_get_latest_ledger_response.json
noleto: Agreed, that’s a good point. This is probably something to address when reorganizing how we test our handlers. On a side note, initially I wrote all the test cases “inline” (generated by Claude). While reviewing them, I noticed that Claude often produces tautological tests. I then generated actual real payloads and asked Claude to write tests based on those payloads. The results were much better they caught real logic issues in my code and provided solid edge-case coverage. So something I'm in favor to be generalized for all handlers

noleto:
> I think this PR probably could have been split into 2 tbh.

I did refactor some code beyond the Stellar requirements, but aside from the `HandlerType` refactoring, everything else was directly tied to what Stellar needed. That said, I see your point  the PRs are still quite verbose, which to me suggests that our abstractions are still too large. Handlers were a good step, but we need to push things further and keep separating concerns so that adding a new handler becomes a straightforward, low-impact change in the codebase. Definitely something to discuss.

lib/network/beacon_handler.go
```diff
@@ -36,8 +36,8 @@ func NewBeaconChainHandler(config *NetworkConfig) *BeaconChainHandler {
 }
 
 // Metadata methods for registry
-func (h *BeaconChainHandler) GetType() string {
-	return "beacon-chain"
+func (h *BeaconChainHandler) GetType() HandlerType {
```
smoyer64, line 39: We can do more of this - excellent work typing these string constants as part of adding a feature 🔥

smoyer64 (approved):

PR #187, opened 2026-02-17
Complete ExtractMethod for Esplora

AusIV:
This implements ExtractMethod for documented esplora paths, reducing them to a single method identifier per path pattern.

### DIFF
--- a/lib/network/bitcoin_esplora_handler.go
+++ b/lib/network/bitcoin_esplora_handler.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


smoyer64 (approved):

cds-amal (commented):
Hey, nice work on this one. The old `return "REST"` was a total black hole for observability; this gives us real per-endpoint visibility without blowing up Prometheus cardinality and leaving us in the dark :). The regex coverage is thorough, the test table is a nice touch, and precompiling the patterns at package level is the right call.

A few suggestions, I strongly urge you to consider:

**Stringly typed returns**: right now the endpoint IDs are raw string literals in the switch and duplicated in the test. If you typo `"GET_BLCOK_TXS"` in one place, the compiler won't catch it. Go has named types with constants, and gives all the fancy compiler and IDE niceties:

```go
type EndpointName string

const (
    GetTx         EndpointName = "GET_TX"
    GetTxStatus   EndpointName = "GET_TX_STATUS"
    // ...
    Unknown       EndpointName = "UNKNOWN"
)
```

Then the switch returns `GetTx` instead of `"GET_TX"`, the test asserts against the same constants, and you get autocomplete plus compile-time typo detection. The conversion to `string` at the `ExtractMethod` boundary is a one-liner. (This also sets us up nicely for the `Operation` type in the refactoring roadmap; these constants map directly to what would become `Operation.Category`.)

**Small things**:

- The `$$` at the end of every regex (e.g., `/tx/[^/]+$$`) is functionally fine but reads like a typo; single `$` does the same thing in Go regex. I had to look that up, it was weird seeing so many of them stare back at me, knowing it compiled.
- The doc comment says "All regexes are anchored (`^...$`)" but there's no `^` anchor on any of them. The suffix-only matching is intentional (and the test proves it works), so the comment should reflect that.
- The switch evaluates top-to-bottom; putting more-specific patterns before their general siblings (e.g., `reBlocksTipHeight` before `reBlocks`) would make the ordering obviously safe rather than subtly safe; we don't want subtly safe.

**Bigger picture**: this commit solves a real problem, and it solves it well for Esplora. But have you looked at what Tron and Beacon are doing right now?

`tron_handler.go:140` and `beacon_handler.go:88` both have the exact same `return req.URL.Path, nil` that Esplora used to have. They're the same metrics dead zone addressed here. So the question is: what if the regex table + match function you built (`EndpointID`) wasn't just an Esplora thing, but a reusable type that each REST handler populates with its own route definitions?

Something worth thinking about: Tron's `GetSupportedMethods()` already lists 86 static paths, organized by API category. Those are basically a route table waiting to happen. Beacon has 9. The data is already there; it's just not connected to `ExtractMethod` yet.

The reason I bring this up is the `NetworkHandler` interface (37 methods and counting). Every REST handler has to implement `ExtractMethod` from scratch, and they all end up solving the same problem: "match an incoming path against known routes, return a stable label." What if instead of each handler reimplementing that logic, there was a small, composable type (regex table + match function) that handlers compose into their struct? That's the ISP pattern in action: behavior provided by composition, not by being arbitrary methods on a giant interface.

This commit is a chance to establish that pattern, not just fix one handler. The `EndpointID` function and the precompiled regex table are already 90% of the abstraction; the remaining 10% is making it generic enough that Tron and Beacon can use the same machinery with their own route definitions. What would that look like?

Overall this is a solid step forward. It turns a metrics dead zone into real observability with a clean, testable design. And if this becomes the template for how all REST handlers do method extraction, it goes from a good fix to a great architectural pattern.

lib/network/bitcoin_esplora_handler.go
```diff
@@ -98,9 +99,172 @@ func (h *BitcoinEsploraHandler) ProcessRequest(req *http.Request) error {
...
+	reAssetsRegistry    = regexp.MustCompile(`/assets/registry$$`)
+)
+
+// EndpointID maps an HTTP method + request path to a stable, unique endpoint identifier.
+//
+// Notes:
+//   - Paths below are derived from Blockstream Esplora API docs.
+//   - `path` should be the URL path only (no scheme/host/querystring), e.g. "/tx/<txid>/status".
+//   - All regexes are anchored (^...$) so partial matches won't collide.
+func EndpointID(method, path string) string {
```
smoyer64, line 161: Thinking more about this after @cds-amal 's comment, it does seem like it would be prudent to type the return of this function so that the output string could not be confused for anything other than an identifier.

smoyer64 (dismissed):

AusIV:
I've corrected the documentation error - In my original iteration they were anchored as documented, but I removed that out of uncertainty to whether prefixes like /bitcoin-esplora/ or provider-specific prefixes might be present at the point that we were checking that, and missed the documentation.

I've also removed the double '$$' - that slipped in when I removed the anchoring, but I agree it's not necessary and just adds clutter.

As far as evaluation order: Because the regular expressions are anchored at the end and not the beginning, the place the order matters most is actually at the beginning of the regex, not the end. When I implemented tests, I initially found that paths intended for `reAssetTxsMempool` inadvertently matched `reMempool`. `reMempool` could also get tripped up  with `reAddressTxsMempool` and `reScripthashTxsMem`, so it is deliberately positioned after all of them. Otherwise I'm open to suggests around alternative orderings, but aside from conflicts with `reMempool` (which are already addressed deliberately) none of the other expressions would match one another's patterns.

I'm a bit hesitant to adopt the strongly typed returns. Having a bunch of EndpointName constants in the network package, where the handlers for every different type of system we support are all in the same package, seems like a recipe for namespace collisions and confusion. And given that `EndpointId()`'s return values are used as the value for `ExtractMethod()` (which usually extracts the method from a JSON RPC), I'm not sure that separating that out as a different type has a lot of benefit. Perhaps also notably, the one place other than metrics that we use this information is for method-based routing, in which the method must be specified as a string in the config, and I'm not sure how we'd map that back to defined constants.

 As far as the more generalizable solution: I agree that something like that would be good, but the sense of urgency I got from Tom on getting visibility into these methods was somewhat at odds with deriving a well generalized solution. I know that there are plans in the near future to aim for getting new API support out in a matter of hours, and I think at that point we need to re-evaluate this kind of network-specific data extraction in terms of config changes instead of code changes, so I'd probably advocate for a more comprehensive evaluation at that point.

natefikru (approved):

smoyer64 (approved):

PR #188, opened 2026-02-19
test(prometheus): simplify HandleRequestMetrics and remove priority l…

smoyer64:
Remove the redundant `requestBody` parameter from `HandleRequestMetrics`
  since the RPC method is already captured in `PromRequestMetricData.Method`,
  eliminating the coupling to `din_http.JSONRPCRequest` across the interface,
  implementation, mock, and all call sites.

  Also remove the `priority` label from `DinProviderHealthCheckCount` to
  reduce metric cardinality.

  Additional improvements:
  - Add `nowFunc` injectable clock to `HybridSampler` for deterministic time-based tests, replacing the previously untestable `time.Now()` call
  - Fix `TestMain` to call `os.Exit(m.Run())` so test failures are reported
  - Add missing `api_key` label assertion in `TestHandleRequestMetric`
  - Add `newTestNowFunc` helper for stable, incrementing timestamps in tests
  - Register prometheus mock generation in the Makefile `generate-mocks` target


lib/prometheus/sampler_test.go
```diff
@@ -313,6 +316,7 @@ func TestShouldSampleHealthCheck(t *testing.T) {
 
 // TestLabelCollisionResistance verifies that similar labels don't cause hash collisions
 func TestLabelCollisionResistance(t *testing.T) {
+	t.Skip()
```
cds-amal, line 319: Ah, the new inject time breaks idempotency where repeated calls with the same labels produce different results as time marches on because the stability tests (same labels = same result), which is incompatible with the time-in-hash design.

Because the time-in-hash design changed the playing field, we should fix it to use `nowFunc` and rewrite the assertions to test what the function name actually says: that similar-but-distinct label sets (like `["ab", "cd"]` vs `["a", "bcd"]`) don't all land on the same sampling decision? That's the poperty the pipe delimiter in the hash is there to guarantee, and it's worth having a test for.

I prefer the test re-written

lib/prometheus/prometheus_test.go
```diff
@@ -202,6 +204,8 @@ func TestHandleHealthCheckMetric(t *testing.T) {
 func TestHandleNetworkHealthCheckMetric(t *testing.T) {
 	// Initialize the prometheus client
 	client := NewPrometheusClient(logger.NewLoggerClient(zap.NewNop(), utils.Environment("test")), "test-machine-id")
+	// client.requestSampler = NewHybridSampler(1.0, 1.0)
```
cds-amal: Experience the mini-joy of deleting this line of code :)

lib/prometheus/prometheus_test.go
```diff
@@ -219,7 +223,7 @@ func TestHandleNetworkHealthCheckMetric(t *testing.T) {
 			data: &PromNetworkHealthCheckMetricData{
 				Network:        "/ethereum",
 				ResponseStatus: 200,
-				Duration:       100 * time.Millisecond,
+				Duration:       90 * time.Millisecond,
```
cds-amal: This looks like boundary discovery experiments that made it into the PR;  If they're necessary, please add a doc-comment, or  revert them so we don't have to stitch context in the future.

edit: s/remove/revert

lib/prometheus/prometheus_test.go
```diff
@@ -229,14 +233,14 @@ func TestHandleNetworkHealthCheckMetric(t *testing.T) {
 				"environment":     "test",
 			},
 			expectedCount:    1,
-			expectedDuration: 100,
+			expectedDuration: 90,
```
cds-amal, line 236: same boundary discovery experiments

lib/prometheus/prometheus_test.go
```diff
@@ -229,14 +233,14 @@ func TestHandleNetworkHealthCheckMetric(t *testing.T) {
...
 		},
 		{
 			name: "Valid Data - Error Status",
 			data: &PromNetworkHealthCheckMetricData{
 				Network:        "/polygon",
 				ResponseStatus: 503,
-				Duration:       50 * time.Millisecond,
+				Duration:       40 * time.Millisecond,
```
cds-amal: same boundary discovery experiments

lib/prometheus/prometheus_test.go
```diff
@@ -246,7 +250,7 @@ func TestHandleNetworkHealthCheckMetric(t *testing.T) {
 				"environment":     "prod",
 			},
 			expectedCount:    1,
-			expectedDuration: 50,
+			expectedDuration: 40,
```
cds-amal, line 254: same boundary discovery experiments

cds-amal (changes requested):
Thanks @smoyer64 ! This turns the lights on for testing. I left a few comments.

lib/prometheus/sampler_test.go
smoyer64: No, that test can't actually do what it's name and messages imply.  I plan to write an issue for that one once this is merged.

lib/prometheus/prometheus_test.go
smoyer64: Kind of, it's me moving the test value away from the potential FP equality test as discussed last evening (4 typ).  IRL, we don't really care whether ~100.0 is bucketed into 50.00000000...1-100.0 or 1.00000000000...1-150.0, but for tests we do, so I moved the test away from the Histogram bucket transitions.  I'll add a note to the lines setting the values anyway.

lib/prometheus/sampler_test.go
cds-amal: Sounds good!

smoyer64: Here's the promised GH Issue describing the problem with this test: https://github.com/DIN-center/din-caddy-plugins/issues/189

cds-amal (approved):
Thanks @smoyer64, good to go. Let's get that fire from Prometheus!

PR #194, opened 2026-02-24
refactor: unify proxy architecture — shared abstractions, module restructure, health checker framework

natefikru:
## Summary

Unifies the network/RPC and AI proxy architectures by extracting shared abstractions, restructuring modules into symmetric sub-packages, and creating a reusable health check framework.

### Phase A: Shared Abstractions
- Extract `HealthStatus` enum into `lib/health/status.go` (eliminates 3 duplicate definitions)
- Extract shared `Provider` interface into `lib/provider/provider.go`
- Both proxies implement the shared `Provider` interface
- Migrate AI proxy from local `HealthStatus` to shared `lib/health.HealthStatus`
- Update prometheus sampler for lowercase health status strings

### Phase B: Module Restructure
- Move 30 network proxy files from `modules/` root to `modules/network/` sub-package
- Symmetric layout: `modules/network/` and `modules/ai/` (no .go files in `modules/` root)
- Single import path change in `module.go` — Caddy module IDs unchanged

### Phase C: Shared Health Checker
- Extract common ticker/quit/overlap-prevention pattern into `lib/health/checker.go`
- Migrate AI health checks to `health.Checker` (with `PreventOverlap: true`)
- Migrate network health checks to `health.Checker` (with `RunImmediately: true`)

### Phase D: Shared Metrics Interface
- Create `metrics.HealthCheckRecorder` interface in `lib/metrics/`
- Implement on network `PrometheusClient` and AI `AIMetricsClient`
- Compile-time interface checks on both implementations

## What does NOT change
- Caddy module IDs (`http.handlers.din`, `http.handlers.din_ai`, etc.)
- Caddyfile syntax or directive names
- Health check logic, thresholds, or state machines
- Metric names or label schemas (no dashboard breakage)
- No behavioral changes to request routing

## New shared packages

| Package | Purpose |
|---------|---------|
| `lib/health/status.go` | `HealthStatus` enum shared by both proxies |
| `lib/health/checker.go` | Reusable periodic health check loop with graceful shutdown |
| `lib/provider/provider.go` | `Provider` interface shared by both proxy types |
| `lib/metrics/metrics.go` | `HealthCheckRecorder` interface for health check metric recording |

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./...` passes (unit tests)
- [x] `go build -tags=integration ./modules/ai/...` passes (integration tests)
- [x] `go vet ./...` passes
- [x] No `.go` files remain in `modules/` root
- [x] Both health checkers use `health.Checker` framework
- [x] Both metrics clients satisfy `metrics.HealthCheckRecorder`
- [x] Documentation updated (AGENTS.md, docs/ai-smart-routing.md, code comments)


lib/health/checker.go
smoyer64, line 1: Here's a very respected health-check library for Go that's been working well for three (?) years in the `ipfs-p2p-server` project: https://pkg.go.dev/github.com/alexliesenfeld/health.  While we certainly don't need it for this project, it's also compatible with the Health Check RFC (which specifies the format of the JSON that's returned from the (e.g.) `/health` end-point.

smoyer64 (changes requested):
I didn't really check the module restructure part of this PR but with respect to the "shared abstractions", I think we need to talk about what a truly abstracted router would look like.  For now I'm going to ignore the fact that Caddy itself is an HTTP router without additional configuration - and I don't know whether the efficiency of TCP/UDP routing has benefits (it certainly does allow us to keep all routing in a single application if so).

But ignoring the Caddy question, is it time to tackle an abstracted `NetworkHandler`?  I've been mulling this over for quite some time as I've learned how more of the specific handlers work, and I'm pretty convinced that an abstracted network handler would be something like:

```go
package network

import "github.com/alexliesenfeld/health"

type Handler[T, R any] interface {
	// Request handling
	Do(request T) (R, error)

	// Health
	Checks() []health.Check

	// === Lifecycle ===
	Initialize(config *ServiceConfig) error
	Close() error
}
```

**This is the current `network.NetworkHandler` with the stutter removed.**

This is not in any way finished, but realistic considering the type of `network.Handler` and the configuration and capabilities of each provider can all be inject via the `Initialize` function.

One other comment about the `Checks` method - this ignores the fact that we want to collect might want to collect certain data (e.g. the block number for blockchain-based services) during a check.  There's nothing that keeps a health-check from having a side-effect (yes, it's bad form in functional-style programming).  This can be side-stepped by having a data collector and having the `Check` simply report on that data.

This also ignores the differing capabilities of each provider (for a given network, perhaps only some support archive).  I almost added `Capabilities(name string) bool` to the above but those can also be controlled by injecting service/provider constraints via `Initialize`.

I started thinking about this more as I skimmed your AI handler PoC - That would be a good testcase for how we'd generalize a `Handler`.  For AI, it would be cool to call the `/v1/models` end-point to populate the "capabilities" (matrix?). To go deeper, I could also look at how we'd implement the four (typical) IPFS services, two of which are TCP/UDP+QUIC but I'd have to spend the time looking at the plugin that makes Caddy TCP/UDP capable.

PR #199, opened 2026-03-02
perf: reduce allocations and GC pressure in ServeHTTP hot path

### LINKED ISSUE #198: perf: reduce allocations and GC pressure in ServeHTTP hot path
## Summary

Profiling the core `ServeHTTP` request pipeline via Go benchmarks and `pprof` (CPU + memory) revealed significant and avoidable allocation pressure on every proxied request. This creates GC churn that shows up as latency spikes under sustained load.

### Benchmark baseline (Apple M4 Pro, arm64)

```
BenchmarkServeHTTP_RPC_HappyPath-14    342588    17272 ns/op    14695 B/op    134 allocs/op
BenchmarkServeHTTP_RPC_LargeBody-14    342919    17650 ns/op    17000 B/op    136 allocs/op
BenchmarkServeHTTP_Parallel-14         303828    19715 ns/op    14694 B/op    134 allocs/op
```

134 allocations and ~14.7 KB allocated per request. At 1000 req/s that is 134k allocs/s and ~14.7 MB/s of short-lived heap churn feeding the GC.

### pprof memory profile (342k+ iterations, ~60 GB cumulative allocations)

| Allocator | Cumulative | % of total | Root cause |
|---|---|---|---|
| `io.ReadAll` | 27.78 GB | 46% | Body buffer reads, multiple internal grows |
| `strings.genSplit` | 7.93 GB | 13% | `strings.Split` on URL path every request |
| `NewResponseWriterWrapper` | 4.07 GB | 6.7% | `new(bytes.Buffer)` starts at 0 cap |
| `bytes.(*Buffer).grow` | 3.34 GB | 5.5% | Response buffer growing from 0 to final size |
| `bytes.NewReader` | 2.45 GB | 4% | Body restoration in retry loop |
| `io.NopCloser` | 0.85 GB | 1.4% | Body wrapping |

### CPU profile

Dominated by GC activity (`runtime.scanobject`, `runtime.gcDrain`, `runtime.mallocgc`) caused directly by the allocation pressure above. This is the mechanism by which allocations become latency.

---

### 1. Body buffer pool (targets 46% of allocations)

**File:** `modules/din_middleware.go:607-613`

`io.ReadAll` uses `bytes.Buffer.ReadFrom` internally, which triggers a series of `grow()` calls (64 -> 128 -> 256 -> ... bytes) before reaching the actual body size. Add a `sync.Pool` of pre-grown `*bytes.Buffer` objects to eliminate these intermediate allocations.

```go
var bodyBufPool = sync.Pool{
    New: func() any {
        b := new(bytes.Buffer)
        b.Grow(4096)
        return b
    },
}
```

Replace `io.ReadAll` with:

```go
buf := bodyBufPool.Get().(*bytes.Buffer)
buf.Reset()
if _, err := buf.ReadFrom(r.Body); err != nil {
    bodyBufPool.Put(buf)
    return nil
}
bodyBytes := make([]byte, buf.Len())
copy(bodyBytes, buf.Bytes())
bodyBufPool.Put(buf)
repl.Set(RequestBodyKey, bodyBytes)
r.Body = io.NopCloser(bytes.NewReader(bodyBytes))
```

The pooled buffer is returned immediately after the copy. The `bodyBytes` slice is a single allocation of exact size with no wasted capacity.

### 2. Path parsing: replace `strings.Split` with `strings.IndexByte` (targets 13%)

**File:** `modules/din_middleware.go:523-526`

Current code allocates a trimmed string plus a `[]string` slice on every single request just to get the first path segment:

```go
fullPath := strings.TrimPrefix(r.URL.Path, "/")
pathSegments := strings.Split(fullPath, "/")
networkPath := pathSegments[0]
```

Replace with zero-allocation equivalent:

```go
path := r.URL.Path
if len(path) > 0 && path[0] == '/' {
    path = path[1:]
}
networkPath := path
if idx := strings.IndexByte(path, '/'); idx >= 0 {
    networkPath = path[:idx]
}
```

`strings.IndexByte` returns an int. String slicing does not allocate. This removes 2 allocs on every request.

### 3. Pre-size `ResponseWriterWrapper` buffer (targets 5.5%)

**File:** `modules/response_writer_wrapper.go:22`

`new(bytes.Buffer)` creates a zero-capacity buffer. The first write triggers the grow cycle (64 -> 128 -> 256 -> 512 bytes) before settling at the response size. Pre-allocate 512 bytes to cover the vast majority of JSON-RPC responses without any growth:

```go
body: bytes.NewBuffer(make([]byte, 0, 512)),
```

Note: pooling this buffer is not safe because `handlePostRequestTasks` and its spawned goroutines access `rww.body.Bytes()` asynchronously after the HTTP response is written. A copy-before-pool pattern would negate the benefit.

### 4. Increase `MaxIdleConnsPerHost` (reduces TCP overhead under concurrent load)

**File:** `lib/http/http.go:57-58`

Current value is 16 for both `MaxIdleConnsPerHost` and `MaxIdleConns`. At 100+ concurrent requests to the same provider the pool is exhausted, forcing new TCP + TLS handshakes which add 20-100ms per request.

```go
MaxIdleConnsPerHost: 100,
MaxIdleConns:        100,
```

---

## Expected outcome

- 40-50% reduction in B/op and allocs/op based on profile attribution
- Reduced GC pause frequency under sustained load
- Eliminated TCP reconnect latency spikes at high concurrency

## Out of scope for this issue

- Streaming pass-through without buffering (requires removing retry support, separate architectural decision)
- Pooling the `ResponseWriterWrapper` struct itself (blocked by async goroutine lifecycle)
- `caddy.Replacer.Set` allocations (inside Caddy internals)
- `handlePostRequestTasks` goroutine overhead

## How to verify

```bash
# Before
go test ./modules/ -run='^$' -bench='BenchmarkServeHTTP' -benchtime=5s -count=3 -benchmem

# Apply changes, then after
go test ./modules/ -run='^$' -bench='BenchmarkServeHTTP' -benchtime=5s -count=3 -benchmem

# Confirm strings.genSplit and bytes.Buffer.grow are gone from profile
go test ./modules/ -run='^$' -bench='Benchmark' -benchtime=5s -memprofile=/tmp/din_mem_after.prof
go tool pprof -text -nodecount=20 /tmp/din_mem_after.prof

# Full suite
go test ./...
```

natefikru:
Closes #198

## Motivation

Live pprof profiling at 20 concurrent connections (5000 requests against ETH mainnet):

**Before:**
- p50: 58ms | p95: 158ms | p99: 313ms | throughput: 242 req/s
- Memory profile top allocators: `io.ReadAll` 46% | `strings.genSplit` 13% | `bytes.Buffer.grow` 5.5%
- CPU profile: 8% of CPU spent in `compress/flate` (gzip decompression)

**After:**
- p50: 38.7ms | p95: 122.7ms | p99: 363ms | throughput: 279 req/s
- p50 improved -33%, p95 improved -22%, throughput +15%
- `io.ReadAll`, `strings.genSplit`, and `bytes.Buffer.grow` are gone from the top 15 memory allocators

## Changes

**1. Body buffer pool** (`modules/din_middleware.go`)

`io.ReadAll` for reading the request body accounted for 46% of cumulative allocations. It internally uses `bytes.Buffer.ReadFrom`, which triggers a grow chain (64->128->256->... bytes) before reaching body size, then throws those intermediate slices away.

Added a package-level `sync.Pool` of pre-grown `*bytes.Buffer` (4096 bytes). On each request: get buffer from pool, `ReadFrom` into it (no grows for typical payloads), `make([]byte, buf.Len())` + `copy` to extract final bytes (one exact-size allocation), return buffer to pool. `io.ReadAll` allocations are eliminated from the hot path.

Also switches `bytes.NewBuffer` to `bytes.NewReader` for body restoration on retry. `NewReader` is a read-only wrapper with no internal write state.

**2. Zero-allocation URL path parsing** (`modules/din_middleware.go`)

`strings.TrimPrefix` + `strings.Split` allocated a `[]string` on every request just to extract the first path segment. Replaced with `strings.Cut` which returns sub-slices of the original string with no allocation. Removes 1 alloc/op unconditionally. `strings.genSplit` is eliminated from the top 15 memory allocators.

**3. MaxIdleConnsPerHost: 16 -> 100** (`lib/http/http.go`)

At 16 idle connections per host, 20 concurrent requests to the same upstream provider exhausts the connection pool. Requests beyond the pool limit trigger new TCP+TLS handshakes. On the chainstack endpoints used in testing, handshakes add 20-40ms per affected request. Increasing to 100 ensures the pool is never exhausted at realistic concurrency levels. This is the primary driver of the p50/p95 latency improvements in the live test results.

**4. Avoid double gzip decompression** (`modules/din_middleware.go`, `modules/helpers.go`)

The main request path decompresses gzip responses at line ~748 to pass decompressed bytes to `ParseResponse` for JSON-RPC error detection. `handlePostRequestTasks` (the async post-processing goroutine) was then reading `rww.body.Bytes()` and decompressing the same bytes a second time for health-check processing.

Added `DecompressedBody []byte` to `PostRequestTaskParams`. The successful attempt caches its decompressed body and passes it through. `handlePostRequestTasks` uses the cached body directly and skips the second decompress. For failure paths where no cached body is available, it falls back to decompressing from `rww.body` as before.

## Benchmark results (micro)

Micro-benchmarks are dominated by `httptest.NewRequest` + gomock overhead and do not reflect real-world latency. They are included as a regression baseline for allocation counts, not ns/op.

```
                             BEFORE                         AFTER
HappyPath    17272 ns  14695 B  134 allocs  ->  ~25000 ns  14330 B  133 allocs
LargeBody    17650 ns  17000 B  136 allocs  ->  ~26000 ns  15424 B  133 allocs  (-9.2% B, -3 allocs)
Parallel     19715 ns  14694 B  134 allocs  ->  ~30000 ns  14238 B  133 allocs
```

ns/op variance is benchmark noise. B/op and allocs/op reductions are real.

## Live load test results

5000 requests, 20 concurrent, `eth_blockNumber` against ETH mainnet (chainstack):

```
              BEFORE    AFTER    DELTA
p10           38ms      32ms     -16%
p50           58ms      38.7ms   -33%
p75           --        48.7ms
p90           --        107ms
p95           158ms     122.7ms  -22%
p99           313ms     363ms    (noise)
Fastest       30ms      27.1ms   -10%
Req/sec       242       279      +15%
```

## Memory profile (top allocators before vs after)

| Allocator | Before | After |
|---|---|---|
| `io.ReadAll` | 27.78 GB (46%) | not in top 15 |
| `strings.genSplit` | 7.93 GB (13%) | not in top 15 |
| `bytes.Buffer.grow` | 3.34 GB (5.5%) | not in top 15 |

## Test plan

- [x] `go test ./...` passes
- [x] Live load test: 5000 requests, 20 concurrent, all 200s
- [x] pprof memory profile confirms target allocators eliminated
- [x] `modules/bench_servehttp_test.go` included as repeatable alloc baseline


modules/din_middleware.go
```diff
@@ -46,6 +46,17 @@ var (
 	// _ caddy.Validator            = (*mod.DinMiddleware)(nil)
 )
 
+// bodyBufPool reuses buffers for reading request bodies, reducing allocations
+// and GC pressure on the hot path. Each buffer is pre-grown to 4 KB which covers
+// typical JSON-RPC payloads without any internal grow calls.
+var bodyBufPool = sync.Pool{
```
smoyer64, line 52: Nice!

smoyer64 (changes requested):
Correct me if I'm wrong, but this work is to optimize the request/response behavior "outside" the processing that occurs within the selected handler.  What opportunities do you see to optimize the individual handlers?  Obviously, the JSON-RPC handler would be a great starting point if there are gains to be made.  I'm also wondering if you subtracted what should be a pretty constant behavior of the mock handler.  That handler is likely fast compared to a "real" handler, but it's still worth knowing that number.

https://github.com/DIN-center/din-caddy-plugins/blob/f092fd36f63b33cd9118f3b7760d6904d23da035/modules/din_upstreams.go#L122-L161

At one point we had talked about the fact that which providers were deemed healthy for a given service (network) is a Slowly Changing Dimension.  While this code is actually executed after the request processing part of the handler, it is executed (for current services) every time a request is received.  This seems like a pretty easy win as we already know when the data needs to be updated (it's that troublesome write-lock.)

Leo had a diagram that I haven't yet found that shows the whole provider selection process for each request - the left ~half of that diagram should be calculated once each time the provider health data is received, and the right ~half (which would later need to include method-based routing) should be performed during each request/response cycle.

modules/din_upstreams.go
```diff
@@ -86,16 +94,42 @@ func (d *DinUpstreams) GetUpstreams(r *http.Request) ([]*reverseproxy.Upstream,
...
+		}
+
+		pool := d.buildUpstreamPool(providers, nil)
+		// Only cache non-empty pools. An empty pool means all providers are unhealthy;
+		// Caddy's circuit breaker may recover before the next health check cycle, so
+		// we must not cache the empty state.
+		if len(pool) > 0 {
+			d.poolCache.Store(net.Name, &cachedPool{version: currentVersion, pool: pool})
+		}
+		return pool, nil
```
smoyer64, line 129: In L127, you don't know that what you calculated has been stored in the cache without reading it back out of the cache.  Theoretically, the value you just calculated is the same (or at least valid), so this should be fine but might not be consistent.

modules/din_upstreams.go
```diff
@@ -86,16 +94,42 @@ func (d *DinUpstreams) GetUpstreams(r *http.Request) ([]*reverseproxy.Upstream,
...
+				return cached.pool, nil
+			}
+		}
+
+		pool := d.buildUpstreamPool(providers, nil)
+		// Only cache non-empty pools. An empty pool means all providers are unhealthy;
+		// Caddy's circuit breaker may recover before the next health check cycle, so
+		// we must not cache the empty state.
+		if len(pool) > 0 {
+			d.poolCache.Store(net.Name, &cachedPool{version: currentVersion, pool: pool})
```
smoyer64, line 127: A new value is added to the cache here, but when should the old value be deleted?  This is a logical place, but the old value might still be in use by another request.  What is the impact of write locks on this cache (for both storing and deleting cache items?

Issue #205, opened 2026-03-10
din_http_request_count under-counts requests by 50%+ — non-retryable error paths skip metrics


noleto:
**Summary**
The `din_http_request_count` Prometheus counter does not record metrics for a large class of valid requests. In production (SigNoz), the counter is off by more than 50% compared to the actual request volume served by DIN router (see [DIN Metric Audit](https://docs.google.com/spreadsheets/d/14wL17bZJH1plBAj49y62c_lgAoy5dzLODe4pb06dG0I/edit?usp=drive_web&ouid=100790945037140232255)). The metric is supposed to be unsampled (every request counted), but it silently drops requests that exit the retry loop via a non-retryable error path.

**Root Cause**
The offending code is under this file: https://github.com/DIN-center/din-caddy-plugins/blob/2db473240dc8d2ed506d44adab809961b63b3a04/modules/din_middleware.go#L518 

`shouldLogMetrics` is only set to `true` in two places:

- Line 735: successful request
- Line 894: HTTP/network transport error on the final attempt

There are three `break` exits from the retry loop that never set` shouldLogMetrics = true`, and when `next.ServeHTTP` returns `nil` (transport succeeded), the fallback block at line 944 (`if err != nil`) is also skipped. The result is a fully served request with no metric emitted.

Path 1 — Non-retryable application error (`din_middleware.go:844-858`)

```
// appError != nil, !IsRetryableError(), !IsRetryableOnDifferentProvider()
logFailedAttempt(...)
break  // ← shouldLogMetrics never set; err == nil after loop → no metric
```
This is the highest-volume path. It covers all valid provider responses for: `eth_call` reverts, invalid params (`-32602`), invalid transaction/block hashes, contract errors, etc. These are extremely common in real blockchain traffic.

**Path 2 — Method-not-found, provider key unavailable (`din_middleware.go:784-797`)**

```
// IsRetryableOnDifferentProvider() = true, but provider key missing from repl
logFailedAttempt(...)
break  // ← shouldLogMetrics never set
```

**Path 3 — All providers excluded (`din_middleware.go:807-821`)**

```
// All providers returned -32601, len(excludedProviders) >= len(providers)
logFailedAttempt(...)
break  // ← shouldLogMetrics never set
```

**Regression History**
[...]

Issue #201, opened 2026-03-03
Report RED metrics


smoyer64:
I learned a lot during my week of front-line DIN support but perhaps the most surprising is that we sometimes don't have metrics that give us the visibility we need to effectively troubleshoot issues.  Since we don't have a separate business dashboard system, we should also send enough data to SigNoz to run the business.

The issue that started my journey into was a report that [INFURA was experiencing rate-limiting on the Monad mainnet](https://consensys.slack.com/archives/C03PPLKSAQ7/p1771409820944479) and yet the dashboards built around metrics showed no errors.  I started to create a traditional [RED](https://grafana.com/files/grafanacon_eu_2018/Tom_Wilkie_GrafanaCon_EU_2018.pdf) (Rate, Errors, Duration) version of the "DIN Overview Dashboard" and soon realized that we simply didn't have the metrics to support graphing errors over time.  This issue describes the problem as well as the proposed changes to send the data we need to our OTEL collector.

## Problem

When a user sends us a request, there are paths that don't record metrics at all, and there are other paths that could (currently) report metrics but don't include the needed labels.  One issue is that there's no convenient way to use one metric for both the customer-facing and provider-facing metrics.  The proposed solution in the next section assumes that the `DinRequestXxx` metrics are each split into two metrics.

This sentence is included in the problem statement instead of the proposed solution because it explains why some of the paths aren't covered.  If we have three requests to providers for a single customer request, and all three fail for different reasons, what do you even report to them?  What do you report in the metrics?  The table below describes the nine paths that I'm aware of surrounding the retry loop.  There are likely more paths before we ever reach this point.  There is also the possibility that the context is cancelled before we call `handleContextCancellation` which appears to only be scoped to the retry loop.

```
┌──────────────────────┬──────────┬──────────────────┬──────────────────────────────┬─────────────────────────────────────────────────────────────────────────────────────┐
│                      │   err    │                  │                              │                                                                                     │
│         Path         │  after   │ shouldLogMetrics │      Metrics recorded?       │                                       Counter                                       │
│                      │   loop   │                  │                              │                                                                                     │
├──────────────────────┼──────────┼──────────────────┼──────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┤
│ 1. Success           │ nil      │ true             │ Yes — via                    │ din_http_request_count with real health status, actual response_status              │
│                      │          │                  │ handlePostRequestTasks       │                                                                                     │
[...]

PR #206, opened 2026-03-11
Fix din_http_request_count under-counting

noleto:
This PR tackles #205 
### Problem

`din_http_request_count` was silently dropping metrics for the majority of non-successful requests. Investigation revealed four exit paths in the `ServeHTTP` retry loop that called `break` without ever recording metrics:

| Path | Condition | Introduced by |
|---|---|---|
| 1 | Non-retryable application error (`ParseResponse` returns error, `IsRetryableError=false`, `IsRetryableOnDifferentProvider=false`) | `ce1900e` (#122) — removed `shouldLogMetrics = true` |
| 2 | Method-not-found (`IsRetryableOnDifferentProvider=true`) but provider key missing from replacer | `dd4f8b3` (#181) |
| 3 | All providers excluded after `-32601` failover (`len(excluded) >= len(providers)`) | `dd4f8b3` (#181) |
| 4 | Transport/network error — `shouldLogMetrics = true` was set, but the `if err != nil` return at the end of the function fired *before* `handlePostRequestTasks` was ever reached | pre-existing |

In production, Path 1 alone accounts for the bulk of traffic: any reverted transaction, invalid params, or application-level error takes this branch. The `shouldLogMetrics` flag approach was brittle by design — every new `break` added to the loop was a potential silent regression.

### Fix

Replaced the `shouldLogMetrics` boolean flag with a simpler, impossible-to-miss invariant:

> **If `next.ServeHTTP` was called, `HandleRequestMetrics` must be called — regardless of the outcome.**

The implementation uses a single `internalRequestIssued bool` set immediately after `next.ServeHTTP(rww, r)`. After the retry loop, `handlePostRequestTasks` is called unconditionally when `internalRequestIssued == true`, before any error-return branch. This collapses the four distinct failure paths into one unified post-loop call, and makes any future `break` addition automatically safe.

Additional cleanup:
- **`PrometheusClient` field widened** from `*prom.PrometheusClient` (concrete) to `prom.IPrometheusClient` (interface) — prerequisite for mock injection in tests, and better encapsulation
- **Removed `testMode` guard** from `handlePostRequestTasks` — metrics now work in test mode too; the existing `PrometheusClient == nil` check is the correct guard
- **Removed `Provider` from `PostRequestTaskParams`** — the field was redundant since the provider is already available in the Caddy replacer, which `handlePostRequestTasks` already holds
- **`handleContextCancellation` parameter** similarly widened to accept the interface

### Tests

Tests were written **first** against the unfixed code to confirm they fail, then the fix was applied to make them pass. The suite covers all four previously-uncounted paths alongside control cases and intentional non-counting cases (early returns before the retry loop):

| Test | Expected |
|---|---|
| Success on first attempt | `HandleRequestMetrics` called once |
| Context cancelled (handled by `handleContextCancellation`) | `HandleRequestMetrics` called once |
| Non-retryable app error — path 1 | `HandleRequestMetrics` called once |
| Retryable errors then non-retryable final — path 1 variant | `HandleRequestMetrics` called once |
| Method-not-found, provider key missing — path 2 | `HandleRequestMetrics` called once |
| All providers excluded — path 3 | `HandleRequestMetrics` called once |
| Transport error on final attempt — path 4 | `HandleRequestMetrics` called once |
| Root path `/` health probe | No metrics |
| Unknown network | No metrics |
| Empty body on RPC network | No metrics |
| Payload too large | No metrics |

# Manual Validation Workflow to Verify This PR

### 1. Create an Infura Free Account

-   Create a new project.
-   Restrict the allowed methods to:
    -   `eth_blockNumber`
    -   `eth_chainId`
    -   `eth_getBlockByNumber`
-   The free tier has a low rate limit, which is suitable for testing
    rate limit behavior.

### 2. Configure a Single Network in the DIN Router

Add the following configuration:

``` hcl
networks {
  eth {
    providers {
      https://mainnet.infura.io/v3/<YOUR_INFURA_PROJECT_ID>
    }
    chain_id 0x1
    healthcheck_interval 60
  }
}
```

------------------------------------------------------------------------

## Scenario 1 -- Reproducing the Issue (Before the Fix)

1.  Checkout the `develop` branch.
2.  Build the Docker image locally.
3.  Start the DIN Router.

### Step A -- Trigger Rate Limit Errors

Execute 100 requests to an allowed method:

``` bash
seq 1 100 | xargs -I{} -P10 curl --silent --location 'http://localhost:8000/eth'   --header 'Din-Provider-Info: true'   --header 'Content-Type: application/json'   --data '{"method": "eth_blockNumber", "params": [], "id": 1, "jsonrpc": "2.0"}'
```

**Expected Result:**
The Infura rate limit should be reached.

------------------------------------------------------------------------

### Step B -- Trigger Forbidden (403) Errors

Execute 100 requests using a method that is not allowed by the Infura
API key:

``` bash
seq 1 100 | xargs -I{} -P10 curl --silent --location 'http://localhost:8000/eth'   --header 'Content-Type: application/json'   --data '{"method":"eth_call","params":[{"to":"0x5555555555555555555555555555555555555555","data":"0x9663f88f"},"latest",{"0x5555555555555555555555555555555555555555":{"code":"0x6080604052348015600e575f80fd5b50600436106026575f3560e01c80639663f88f14602a575b5f80fd5b60306044565b604051603b91906061565b60405180910390f35b5f43905090565b5f819050919050565b605b81604b565b82525050565b5f60208201905060725f8301846054565b9291505056fea264697066735822122050547252db97964b6f1b4d4d169979311e144946f2d7e5c045236145e713279764736f6c634300081a0033"}}],"id":2100008201,"jsonrpc":"2.0"}'
```

**Expected Result:**
All responses should return HTTP 403 (Forbidden).

------------------------------------------------------------------------

## Metrics Verification

Open the local Prometheus instance:

http://localhost:9090/query

Review the request counter metrics. Before the fix, the error series may not correctly distinguish between
different error types:
Screenshot from 2026-03-11 22-09-53

------------------------------------------------------------------------

## Scenario 2 -- Validation After the Fix

1.  Checkout branch:

        chore/fix-logic-requests-counter

2.  Stop any previously running containers:

    ``` bash
    docker compose down -v
    ```

3.  Rebuild the Docker image.

4.  Start the DIN Router again.

5.  Repeat Step A and Step B from Scenario 1.

------------------------------------------------------------------------

### Expected Result After the Fix

In Prometheus, the metrics should correctly display distinct time series
for all expected error categories, including:

-   Rate limit errors
-   Forbidden (403) errors

The request counters should accurately reflect the different response types:
Screenshot from 2026-03-11 22-26-38


natefikru (approved):

smoyer64 (approved):
Please correct me if I'm wrong:
1. This will record a metric for every customer request.
2. This will write a log record for every request to a provider (assuming a retry happens)
3. This seems to cover everything I noted in https://github.com/DIN-center/din-caddy-plugins/issues/201 except for the context cancellations.

noleto:
> 1. This will record a metric for every customer request.

Indeed, and to clarify: Every single customer request THAT hits an endpoint (for example, early error such as network doesn't exist or body is empty is not counted)

> 2. This will write a log record for every request to a provider (assuming a retry happens)

This will emit a `metric` increment  for every request to a provider (assuming a retry happens)

> 3. This seems to cover everything I noted in #201 except for the context cancellations.

Context cancellations are counted in this path:
https://github.com/DIN-center/din-caddy-plugins/blob/chore/fix-logic-requests-counter/modules/din_middleware.go#L688-L691

